PR #1, opened 2015-01-15
Port to Python 3

tradej:
Tests pass, some basic operations that I tried work as well. With this SPEC, the package built for me in Rawhide Mock all right, and worked as well.

I have taken the liberty to sort the output lists in ksverdiff as I believe it makes more sense to have the differences alphabetically ordered. 

Commits are rebased so that they should make sense individually, therefore I recommend viewing them one by one (the only exception is the SPEC file overhaul in commit 53c9def, that one could arguably be stand-alone).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,6 +11,8 @@ PREFIX=/usr
 
 TESTSUITE:=tests/baseclass.py
 
+PYTHON?=python
+
 all:
 	$(MAKE) -C po
 
@@ -27,7 +29,7 @@ check:
 
 test:
 	@echo "*** Running unittests ***"
-	PYTHONPATH=. python $(TESTSUITE) -v
+	PYTHONPATH=. $(PYTHON) $(TESTSUITE) -v
 
 coverage:
 	@which coverage || (echo "*** Please install python-coverage ***"; exit 2)
@@ -38,10 +40,10 @@ coverage:
 clean:
 	-rm *.tar.gz pykickstart/*.pyc pykickstart/*/*.pyc tests/*.pyc tests/*/*.pyc docs/kickstart-docs.txt docs/programmers-guide
 	$(MAKE) -C po clean
-	python setup.py -q clean --all
+	$(PYTHON) setup.py -q clean --all
 
 install:
-	python setup.py install --root=$(DESTDIR)
+	$(PYTHON) setup.py install --root=$(DESTDIR)
 	$(MAKE) -C po install
 
 tag:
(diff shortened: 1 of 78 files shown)


tradej:
TIL that if you force push (to acommodate the request for referencing rhbz's), all comments to a PR will get deleted. I am really sorry for that.

Anyhow, what remains to be solved: 
* I am highly in favor of keeping `pykickstart-translations` as a separate package because the content is identical for both `python2-pykickstart` and `python3-pykickstart` subpackages, and doing it other ways would, IMHO, amount to an ugly hack.
* Catching exceptions with the `as` statement: AFAICT, there is no way to do this in a nice way, only by using `sys.exc_info()`, which is really ugly. I recommend releasing a major version and discontinuing dependents.
* I will shortly post a fix re-introducing `make check` as mandatory.

vpodzime:
Other than the inline and per-commit comments these all look good to me. Just please merge the patches a little bit so that it's not like 'fix strip()', 'really fix strip()' and so on.

clumens:
I definitely appreciate all this hard work.  With the code changes looking pretty good, I think the only thing left for me is to solve the translations thing - I just cannot approve having a package for just translations.  Other python packages must have found a way to take care of this, because I don't see a lot of translation packages in repoquery results.  Any idea what they do?

tradej:
@clumens E. g. Django does it like this:

clumens:
Yeah, understood.  I am seeing the same thing with installing into somewhere other than _datadir in other packages too.  I don't especially like that method, but I do like it a lot more than more packages.  I am wondering if it'd be possible to merge everything else you've done and then me deal with the translation thing separately.  It looks like it's potentially going to require me to make a lot of changes to my build process, and I'd really hate to hold you up while I look at that.  Also, I don't want to create merge conflicts for you.

tradej:
Thanks a lot for you consideration. However, I am afraid that since we can't simply convert pykickstart to Python 3 and abandon all the Python 2-based dependents, we definitely need the `python{,3}-pykickstart` subpackages at least. Their existence inevitably leads to this problem.

I think reducing the number of subpackages can be done by putting the binaries into the `python3-pykickstart` package and removing the `pykickstart` package entirely (but that, from the user standpoint, seems hard to understand to me). If you want to cut the `translations` package though, the build time changes must be done, I'm afraid.

clumens:
Okay, what about just dropping the packaging portions of translations entirely?  Just don't package them up in your patches and don't have a subpackage.  I'll take care of the rest here.  I'm not going to do a build without translation support, and I think I have all the code required here to do it correctly.  You've already done the most tedious bit of it by removing all those redefinitions of _.

tradej:
Closing this PR is now obsolete, see referencing PR and bug.

PR #2, opened 2015-02-13
Replaced URLGrabber with requests (rhbz#1141245)

tradej:
Because urlgrabber is being dropped by many significant software projects in
Fedora, Pykickstart can't keep depending on it once it is converted to Python 3
(rhbz#985310).

This commit replaces the dep with requests with SSL verification turned on by
default. This could pose trouble in Python 2 if proper underlying packages
aren't installed [1]. In case this becomes a problem, the SSL_VERIFY constant
in constants.py may be switched to False.

[1] https://github.com/kennethreitz/requests/issues/749#issuecomment-19187417

### DIFF
--- a/pykickstart.spec
+++ b/pykickstart.spec
@@ -15,11 +15,11 @@ Group: System Environment/Libraries
 BuildArch: noarch
 BuildRoot: %{_tmppath}/%{name}-%{version}-%{release}-root-%(%{__id_u} -n)
 BuildRequires: python-devel, gettext, python-setuptools
-BuildRequires: python-urlgrabber
+BuildRequires: python-requests
 %if ! 0%{?rhel}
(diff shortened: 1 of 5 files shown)


tradej:
Sorry, I forgot to incorporate changes in `version.py` into this commit, I'll fix that shortly.

clumens:
On the python2 case, can we somehow detect if the support is there and toggle SSL_VERIFY ourselves?

tradej:
@clumens Disabling SSL verification automatically is IMHO a very bad idea. I can modify the error message to advise the user to check this if the lib is run under PY2, but I'd generally advise against disabling it.

tradej:
Fixed issues and pushed an amended commit.

PR #5, opened 2015-03-26
Python 3 compatibility issues fixes

vpodzime:
Python 3's ``copy.copy()`` works in a different way and hits more issues caused by
attributes accessed in the ``__getattr__`` override/redefinition. Long story short,
``copy.copy()`` creates a new empty object without calling ``__init__``  which it then
tries to populate with stuff from the copied object and as part of that calls
``hasattr()`` which in Python 3 no longer catches infinite recursion errors. The
solution is to raise ``AttributeError`` for all attributes defined in ``__init__`` that
are accessed from the ``__getattr__`` method override.

See http://nedbatchelder.com/blog/201010/surprising_getattr_recursion.html for
more information. Also kudos to bkabrda for nailing this down!

### DIFF
--- a/pykickstart/commands/method.py
+++ b/pykickstart/commands/method.py
@@ -93,6 +93,9 @@ class F19_Method(FC3_Method):
     _methods = FC3_Method._methods + ["liveimg"]
 
     def __getattr__(self, name):
+        if name == "handler":
+            raise AttributeError()
+
         if self.handler.liveimg.seen:
             if name == "method":
                 return "liveimg"


vpodzime:
A bit more has to be done for proper Python 3 support. Adding two more patches to the master-python3 branch.

clumens:
Having looked through the python documentation about __hash__ methods, it looks like what it's telling me is that we should not have __hash__ methods on any objects that have mutable values (which is basically all of them in pykickstart) and that anywhere we are using these objects as hash keys is a bug.  I definitely hate to suggest scrapping a lot of hard (and tedious) work, but I just don't think we can reconcile the fact that these objects are mutable with the fact that changing them will result in the hash being wrong.

vpodzime:
I understand the concerns, but a different angle of view is that there is different behaviour of pykickstart objects on Python 2 and Python 3 because on Python 2 they are all hashable. If we all agree (and I have no trouble doing so) that hashing them even on Python 2 is a bug then we have two choices:
1) implement hash functions that provide hash of the current state of the object
2) not implement hash functions

While 2) is definitely going to result in tracebacks, 1) would give people time to hunt these places down and fix them before we really remove the ``__hash__`` functions (and I don't really mind scrapping the tedious work).

Generally I think that having the ability to use pykickstart objects in places where they require to provide ``__hash__`` on users own risk with the warning that the hash is going to change if the object is modified, is not terribly wrong.

clumens:
I can tell you that it was never my intention that any of these objects be hashable.  I'd never given it any thought at all, in fact, but it seems pretty weird to be able to use a pykickstart object as a hash key.  I'd say that knowing this now, the fact that it's possible to do doesn't necessarily mean that it's something that people ought to do.

Different behavior between python 2 and 3 is definitely a problem, though.  I don't know that anyone is relying on this behavior outside of anaconda.  I did some tests earlier today and didn't find anywhere in pykickstart itself we were relying on using objects as keys.

Now that we are on pykickstart-2.0, I don't mind breaking things like this that were only working incidentally before.  I suppose a migration step of having hash functions for now is the kinder way to proceed, though I worry that (1) no one's going to do anything until they are forced to by making them errors, and (2) I'll just forget to do the removal at some point.

vpodzime:
I'm absolutely okay with dropping the last two patches from this PR. We would then need one more anaconda patch from Python 3 stuff, but that's easy to do.

clumens:
If you do that, I'll merge this request.  I don't think having the objects be non-hashable is going to cause problems outside of anaconda.  Thanks!

vpodzime:
> If you do that, I'll merge this request. I don't think having the objects be non-hashable is going to cause problems outside of anaconda. Thanks!

Done and [pushed changes to our work-in-progress python 3 branch too] (https://github.com/M4rtinK/anaconda/commit/cc2e86e218ec5b8f78f1e847d31c102147a666c0)

PR #6, opened 2015-03-27
Who let the docs out?!

vpodzime:
A piece of "Friday work". Basically I ran pandoc on our Anaconda/Kickstart wiki, editted the result to resolve some minor issues and saved the final result as docs/kickstart_docs.rst. You can check how the docs look like by visitting the file in github's web and letting it render it for you nicely [1]. In order to review, I think the best input should be a diff between what I got from pandoc (should be the same as on the wiki) and the result [2]. See the commit message for a rationale behind this.

[1] https://github.com/vpodzime/pykickstart/blob/master-docs/docs/kickstart_docs.rst
[2] https://vpodzime.fedorapeople.org/pandoc_edited.diff

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,6 @@ po-pull:
 	tx pull $(TX_PULL_ARGS)
 
 docs:
-	curl -A "pykickstart-build" -o docs/kickstart-docs.txt "https://fedoraproject.org/w/index.php?title=Anaconda/Kickstart&action=raw"
 	curl -A "programmers-guide" -o docs/programmers-guide "https://fedoraproject.org/w/index.php?title=PykickstartIntro&action=raw"
 
 check:
@@ -38,7 +37,7 @@ coverage:
 	PYTHONPATH=. coverage report --show-missing --include='pykickstart/*'
 
 clean:
-	-rm *.tar.gz pykickstart/*.pyc pykickstart/*/*.pyc tests/*.pyc tests/*/*.pyc docs/kickstart-docs.txt docs/programmers-guide
+	-rm *.tar.gz pykickstart/*.pyc pykickstart/*/*.pyc tests/*.pyc tests/*/*.pyc docs/kickstart-docs.rst docs/programmers-guide
 	$(MAKE) -C po clean
 	$(PYTHON) setup.py -q clean --all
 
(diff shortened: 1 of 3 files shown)


vpodzime:
It this gets accepted, I think we should cherry-pick the change to other active branches too and possibly also made the docs up to date regarding each particular branch. Of course, it would be absolutely wonderful if we could at some point generate the biggest part of these docs -- the commands and options descriptions -- directly from the sources.

M4rtinK:
That's a really nice idea! :) This should help fix the issue of the docs and Pykickstart code getting out of sync as documentation updates can be submitted with code changes. :)

How hard would it be to also generate content for readthedocs and/or a man page from this ? :)

clumens:
I believe Will is working on (or at least thinking about) having docs generated from the code.

clumens:
Regarding this PR - the main reason I'm doing it the way it is today (docs on the wiki, package grabs them during build) is so there's only one source and the wiki always reflects the latest.  So if I were to merge this, I would definitely want to be able to generate and push docs to somewhere at build time.  readthedocs would suffice, and then I could just update the Fedora wiki to point there.

vpodzime:
> Regarding this PR - the main reason I'm doing it the way it is today (docs on the wiki, package grabs them during build) is so there's only one source and the wiki always reflects the latest. So if I were to merge this, I would definitely want to be able to generate and push docs to somewhere at build time. readthedocs would suffice, and then I could just update the Fedora wiki to point there.

Well, that's basically what I'm trying to do here except that it's the other way around -- instead of fetching the contents of the wiki during build, the sources of the docs are already kept in the git repository and by use of GitHub's rendering of ReST we get nice, user-friendly page that reflects the newest since it actually is the newest. Try having a look at the [GH rendering of the docs from this patch] (https://github.com/vpodzime/pykickstart/blob/master-docs/docs/kickstart_docs.rst). It could get some polishing, but I think it's really nice and absolutely up-to-date. Benefits are:
  * we can enforce docs updates just like we enforce test updates on changes
  * we have the docs under our control -- instead of pulling something potentially random from a wiki anybody with a FAS account could have changed to anything just few minutes ago (especially on rhel7-branch this would be really disappointing)
  * we can have different versions of the docs for different branches (e.g. master vs. rhel7-branch)
  * doing changes in the docs equals sending us patches or submitting PRs and doesn't require FAS account
  * ReST is generally considered easier to write than MediaWiki syntax and is definitely nicer in the raw format

Editting is really simple since GH provides [direct editting of files in the repository] (https://github.com/vpodzime/pykickstart/edit/master-docs/docs/kickstart_docs.rst) (with *Save* being *Create pull request*) and with preview of the changes (' rendering).

We can definitely add pushing the ReST file to readthedocs to one of the ``archive`` or ``tag`` targets of the Makefile.

vojtechtrefny:
I think this is a good idea. I really hate downloading from third party sources during build (yes, mostly because of Trasifex and Zanata, but who knows what could happen with Fedora Wiki in the future...).

Different versions of docs for different versions/branches sounds also very good.

+ 1 for readthedocs -- it is good source for users and it would be nice if we could (automatically) upload it there.

clumens:
Vratislav - Oh, I get it now.  I didn't realize that github just renders the .rst files automatically for you.  I'm going to merge this right now then.  Thanks for doing this!

We'll still want to do something to the fedora wiki to redirect people, perhaps simply a table linking a Fedora version to the right place in git.

bcl:
Note that when you link readthedocs to your project it will automatically update the docs when new commits are made. No need to do anything manually.

PR #8, opened 2015-04-16
Add --kexec flag to reboot

bcl:
This will be used to reboot the installed system using kexec, skipping
the BIOS/Firmware and bootloader to speed things up.


tests/commands/reboot.py
```diff
@@ -64,5 +64,15 @@ def runTest(self):
...
         self.assertEqual(str(cmd), "# Halt after installation\nhalt --eject\n")
 
+class F23_TestCase(F18_TestCase):
+    def runTest(self):
+        F18_TestCase.runTest(self)
+
+        # pass
+        cmd = self.assert_parse("reboot --kexec")
+        self.assertEqual(cmd.kexec, True)
+        self.assertEqual(str(cmd), "# Reboot after installation\nreboot --kexec\n")
```
vpodzime: A test for both ``--eject`` and ``--kexec`` would be nice so that we are sure there are no extra newlines in between the options (I know there aren't because of ``rstrip()``)

vpodzime:
Other than the minor suggestion above this looks good to me.

PR #10, opened 2015-04-20
Add %pre-install section to be used after mounting filesystems

bcl:
This adds support for a %pre-install section which will be executed
after all the filesystems have been setup and mounted. It runs without
chroot.


pykickstart/parser.py
```diff
@@ -193,10 +193,12 @@ def __str__(self):
             retval += '\n%post'
         elif self.type == constants.KS_SCRIPT_TRACEBACK:
             retval += '\n%traceback'
+        elif self.type == constants.KS_SCRIPT_PREINSTALL:
+            retval += '\n%pre-install'
 
         if self.interp != "/bin/sh" and self.interp != "":
             retval += " --interpreter=%s" % self.interp
-        if self.type == constants.KS_SCRIPT_POST and not self.inChroot:
+        if self.type in [constants.KS_SCRIPT_POST, constants.KS_SCRIPT_PREINSTALL] and not self.inChroot:
```
vpodzime: This seems to be not aligned with ``Like %pre these scripts do not run in the chrooted environment.``.

vpodzime:
Generally, this is defitinitely a good idea. But I'm a little bit concerned if it's wise to add users more space for doing crazy things in kickstarts (thinking about the insane ``%pre`` and ``%post`` scripts we have all seen). That said, I'd really like to know what is the motivation behind this before I give it an ACK.

clumens:
Do you see a need for running scripts in other places in anaconda in the future?  If so, I'd like to see a more generic name for the section header, perhaps with some argument specifying where it will be run.  I don't really want to see us add a whole pile of sections, one at a time, over the next year or two.

clumens:
We talked about this a bit in IRC and determined that this is really the only place left in anaconda that's not easily exposed for user scripts.  Plus, giving it the unique name keeps consistency with %pre, %post, and %traceback.  So that's no longer a concern from me.

pykickstart/parser.py
bcl: oops, good catch. I originally had it supporting chroot, until I realized that wouldn't work :)

PR #12, opened 2015-05-20
Update kickstart-docs.rst

jkonecny12:
Fix link to ksdevice to anaconda repository.


vpodzime:
Looks good to me, but we should also remove the ``ksdevice`` docs from the kickstart documentation -- it doesn't belong there and it also has an invalid link in it.

jkonecny12:
If it's not specifically related to this commit I think you can give me Ack a we may create new PR for the ``ksdevice`` part changes.

vpodzime:
That's a good point. ACK.

jkonecny12:
Pushed

PR #14, opened 2015-06-02
Check if python-nose is installed and abort with error if not

atodorov:
This is a simple dependency check.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,6 +39,7 @@ check:
 	tests/pylint/runpylint.py
 
 test:
+	@which nosetests || (echo "*** Please install nosetest (python-nose) ***"; exit 2)
 	@echo "*** Running unittests ***"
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -39,6 +39,7 @@ check:
 	tests/pylint/runpylint.py
 
 test:
+	@which nosetests || (echo "*** Please install python-nose ***"; exit 2)
```
vpodzime: I'd rather see this suggesting users to install ``nosetest``, possibly with the package name in the parentheses. People could try using pykickstart on other distributions too.

atodorov: In this case we'd need to change the messages for coverage and po-pull targets as well.

vpodzime: Let's do it then. Possibly in a separate commit.

PR #19, opened 2015-06-03
Add extra test coverage for commands/btrfs.py

atodorov:


atodorov:
updated with code comments

PR #20, opened 2015-06-03
Improve test coverage for version.py

atodorov:
1) Improve test coverage
2) The new test at line 215 actually revealed a bug, the local variable found wasn't defined and in case of error the finally block raised an exception.

### DIFF
--- a/tests/version.py
+++ b/tests/version.py
@@ -1,4 +1,5 @@
 import six
+import sys
 import unittest
 import tempfile
 import os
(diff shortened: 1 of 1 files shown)


atodorov:
ping. can anyone review and comment on this one ?

pykickstart/version.py
```diff
@@ -178,6 +178,7 @@ def returnClassForVersion(version=DEVEL):
 
     module = module.lower()
 
+    found = ()
     try:
         import pykickstart.handlers
         sys.path.extend(pykickstart.handlers.__path__)
```
vpodzime: This issue is already resolved on master by commit c1fcb70a62ef.

tests/version.py
```diff
@@ -287,5 +298,31 @@ def write_ks_cfg(buf):
...
+            cls = ver.returnClassForVersion(ver.F23)
+
+            # assert the names; b/c of how the imp.load_module() works
+            # asserting both classes being equal fails
+            self.assertEqual(cls.__name__, F23Handler.__name__)
+        finally:
+            # force import to reload these modules
+            del sys.modules['pykickstart.version']
+            del sys.modules['imp']
+            del sys.modules['imputil']
```
vpodzime: I think this needs to import the modules back because tests that run after this one may need it (and if they are in the same module, they wouldn't do the import on their own, I think).

atodorov: I don't think this is necessary. In my test executions this test case is running somewhere in the middle and there are others (from the same module) running after it without any issues.

PR #23, opened 2015-06-10
Install the python3 .mo files to python3_sitelib

dashea:
The .mo files in python3-kickstart are being installed to /usr/lib/python2.7/site-packages, which I don't think is what you want.


vpodzime:
ACK.

PR #24, opened 2015-06-11
Fix bug with adding self.gid

phracek:
Signed-off-by: Petr Hracek <phracek@redhat.com>

I guess that this is a typo.
I have fixed that via pull request.


mulkieran:
Thanks for the patch.

It's not obvious to me how it improves on the original, though.

Can you explain what problem you are fixing?

clumens:
I assume this is happening because self.gid is a string instead of an int?  How are you arriving at that situation?  Can you post a test case?

PR #25, opened 2015-06-13
Add --kexec flag to reboot (#1207747)

bcl:
This will be used to reboot the installed system using kexec, skipping
the BIOS/Firmware and bootloader to speed things up.

Related: rhbz#1207747
(cherry picked from commit 79b4c6fde5fd4e89652ee0b1c2d40bf57f014250)


pykickstart/commands/reboot.py
```diff
@@ -102,3 +102,26 @@ def parse(self, args):
...
+        if self.kexec:
+            retval += " --kexec"
+
+        return retval + "\n"
+
+    def _getParser(self):
+        op = F18_Reboot._getParser(self)
+        op.add_option("--kexec", dest="kexec", action="store_true",
+                      default=False)
+        return op
```
vpodzime: If I understand this correctly, we also allow things like ``poweroff --kexec``, ``halt --kexec`` and similar. Is it what we want or should we also override ``parse()`` here and check that ``--kexec`` is only used in combination with ``reboot``?

bcl:
Strictly speaking --kexec only makes sense with reboot. As it is written it is allowed with all the variations. Since this is a backport of an existing feature I'd rather change/discuss that for master, not the backport.

vpodzime:
Works for me.

vpodzime:
ACK

PR #26, opened 2015-06-17
Add options for LVM cache specs to the 'logvol' command

vpodzime:
As the first step towards the LVM cache support we need to teach the parser to
accept the options for specifying cache size, PVs it should used and mode it
should use.


docs/kickstart-docs.rst
```diff
@@ -1047,6 +1047,19 @@ Create a logical volume for Logical Volume Management (LVM).
     ``--size=``. This option must be used with
     ``--useexisting --size=``, or an error will be raised.
 
+``--cachesize``
+
+    Requested size of cache attached to the logical volume. (Requires --cachepvs)
```
bcl: What units does it expect?

pykickstart/handlers/f23.py
```diff
@@ -56,7 +56,7 @@ class F23Handler(BaseHandler):
         "lang": commands.lang.F19_Lang,
         "liveimg": commands.liveimg.F19_Liveimg,
         "logging": commands.logging.FC6_Logging,
-        "logvol": commands.logvol.F20_LogVol,
```
bcl: Odd. So F21_LogVol wasn't used in F21?

docs/kickstart-docs.rst
vpodzime: Good catch, thanks! It's megabytes as all our storage sizes, but I'm definitely going to make it explicit in the docs.

pykickstart/handlers/f23.py
vpodzime: *f21.py* has these set to ``F21_LogVol`` and ``F21_LogVolData``. But *f22.py* doesn't, so...oops! We should inspect these files and make sure the most recent versions are used in the future.

clumens:
Merged by hand, thanks.

PR #27, opened 2015-06-17
Avoid traceback in module loading failure paths.

dwlehman:
I hit a traceback about `found` being undefined or similar when I tried to run the test suite with a syntax error in my new code.

### DIFF
--- a/pykickstart/version.py
+++ b/pykickstart/version.py
@@ -188,6 +188,7 @@ def returnClassForVersion(version=DEVEL):
             if k.lower().endswith("%shandler" % module):
                 return v
     except:
+        found = None
         raise KickstartVersionError(_("Unsupported version specified: %s") % version)
     finally: # Closing opened files in imp.load_module
         if found and len(found) > 0:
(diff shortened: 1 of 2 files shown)


mulkieran:
found isn't actually a list, probably better to do something like:

```
diff --git a/pykickstart/version.py b/pykickstart/version.py
index 7e3e4b7..33d2a24 100644
--- a/pykickstart/version.py
+++ b/pykickstart/version.py
@@ -178,11 +178,12 @@ def returnClassForVersion(version=DEVEL):

     module = module.lower()

+    fobj = None
     try:
         import pykickstart.handlers
         sys.path.extend(pykickstart.handlers.__path__)
-        found = imp.find_module(module)
-        loaded = imp.load_module(module, found[0], found[1], found[2])
+        (fobj, pathname, description) = imp.find_module(module)
+        loaded = imp.load_module(module, fobj, pathname, description)

         for (k, v) in list(loaded.__dict__.items()):
             if k.lower().endswith("%shandler" % module):
@@ -190,8 +191,8 @@ def returnClassForVersion(version=DEVEL):
     except:
         raise KickstartVersionError(_("Unsupported version specified: %s") % version)
     finally: # Closing opened files in imp.load_module
-        if found and len(found) > 0:
-            found[0].close()
+        if fobj:
+            fobj.close()

 def makeVersion(version=DEVEL):
     """Return a new instance of the syntax handler for version.  version can be
```
and same with other method.

PR #29, opened 2015-06-24
autopart root snapshot support

dwlehman:
The rhel7-branch version is only different in that it omits the F23 specifics.

Related: rhbz#1113207


clumens:
Okay, how do you want to handle the needs docs label, then?  Do you want to add them to this PR or do it later?  Or do you want me to write something?

dwlehman:
Just pushed a new version with an entry for the new option in the docs.

clumens:
Go ahead and push to whichever branches are necessary.  It looks like the autopart object needs to start using getArgsAsStr, but I will do that conversion myself.

dwlehman:
I'm putting this on the shelf indefinitely.

PR #30, opened 2015-07-24
Remove documentation compilation warnings

jkonecny12:
This will remove another two warnings when building documentation on anaconda.

### DIFF
--- a/pykickstart/parser.py
+++ b/pykickstart/parser.py
@@ -161,18 +161,24 @@ class Script(KickstartObject):
     def __init__(self, script, *args , **kwargs):
         """Create a new Script instance.  Instance attributes:
 
-           errorOnFail -- If execution of the script fails, should anaconda
-                          stop, display an error, and then reboot without
-                          running any other scripts?
-           inChroot    -- Does the script execute in anaconda's chroot
-                          environment or not?
-           interp      -- The program that should be used to interpret this
-                          script.
-           lineno      -- The line number this script starts on.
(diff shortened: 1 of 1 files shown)


vpodzime:
The new version is really long. Is there any chance for some more compact form Sphinx would still be happy?

jkonecny12:
I will try something to create more compact version.

jkonecny12:
I tried do my best I hope this is enough.

vpodzime:
Looks better to me, but I'm leaving this to @clumens to judge.

clumens:
I don't know what kind of errors you are seeing.  I don't particularly mind the formatting of documentation one way or the other.  However, there's a lot more docstrings in pykickstart and those should probably all be in the same format, whichever one that is.

bcl:
blivet is using keyword to document kwargs and the results look pretty good, including them in the Parameters block. eg.  :keyword clearPartType: overrides :attr:`self.config.clearPartType`

jkonecny12:
@clumens yeah it's little harder to explain. You will see this compilation error when building docs on anaconda and it's only here because the ``AnacondaKSScript`` class (this class gets the warnings) doesn't have ``__init__()`` method so the sphinx take it from parent which is here.

@bcl I will look on that, thank you for pointing me on that

jkonecny12:
The `keyword` is exactly what I was looking for. Thank you all for valuable input.

jkonecny12:
Pushed. Thank you

PR #31, opened 2015-08-04
Fix liveimg equality check

bcl:
The other object may be None, test for that first before trying to
access potentially non-existant attributes on it.


vpodzime:
Looks good to me.

PR #32, opened 2015-08-07
Add --gpgkey option to repo command

marmarek:
This could be used to verify packages by pungi or livecd-creator, which
have no other way to get key location (no real .repo files).


cgwalters:
I like the idea, but currently we're just wrapping an arbitrary string.  I would hard require this to be a fingerprint, because a lot of people are unaware of how easy it is to create key ID collisions: http://security.stackexchange.com/questions/74009/what-is-an-openpgp-key-id-collision

marmarek:
Actually I was thinking about path to the exported key - same as in
repository definition, not the KEY ID.

Take a look at the pungi patch, which makes use of this option:
https://lists.fedoraproject.org/pipermail/buildsys/2015-August/004852.html

-- 
Best Regards,
Marek Marczykowski-Górecki
Invisible Things Lab
A: Because it messes up the order in which people normally read text.
Q: Why is top-posting such a bad thing?

cgwalters:
Ah, I forgot that the way yum repo files specify keys doesn't support this, just URLs.  Which actually makes me wonder how it works today, as the default for yum repos is supposed to be `gpgverify=1`, but I don't see anything in the yum/dnf payload code setting that to `0`.  Are you sure with this provided GPG verification will actually be used?

cgwalters:
Ok right I remember now.  Yum has this seriously weird thing going on where the default in the *code* is `gpgcheck = 0`, but that's overridden normally by the system `/etc/yum.conf` which has `gpgcheck=1` which applies to all repos.

So for any other project that *isn't* using the global yum conf, they'll need to be careful to flip the default back to `1`.

cgwalters:
(Which I see you're doing in the pungi patch, OK)

cgwalters:
I'm not actually a pykickstart maintainer, but I'm guessing they'd want a test in `tests/commands.repo.py`.  Other than that :+1: from me FWIW.

vpodzime:
Exactly as @cgwalters says -- could you please add a test for the new option to ``tests/commands/repo.py``? Thanks for the patch!

bcl:
Currently gpg signatures are not checked by Anaconda. Would adding this be expected to turn on signature checking for the repo (and write that to the repo config)? I'd like to see option use be consistent across kickstart library users otherwise things will get confusing so a related anaconda patch would be good.

Also please update the documentation in kickstart-docs.rst

marmarek:
Added test and documentation. Working now on support in livecd-tools (almost done - needs some testing). Regarding anaconda support - I'm stuck with fc20 there, so have no easy way to test on recent version. Will see how much it differs.

marmarek:
Offtopic: is the buildsys ML the right place for pungi patches? I've got no response there.

vpodzime:
> Regarding anaconda support - I'm stuck with fc20 there, so have no easy way to test on recent version. Will see how much it differs.

Anaconda now uses DNF for package installation.

These two patches look good to me, but I agree with @bcl that we need to support GPG verification in all places or nowhere. So until there's Anaconda support for this new option, I'm not for merging this.

vpodzime:
> Offtopic: is the buildsys ML the right place for pungi patches? I've got no response there.

I was told that the best way is to provide a pull request at https://pagure.io/pungi

marmarek:
For the record: Pungi patch: https://pagure.io/pungi/pull-request/63
And Anaconda pull request is here: https://github.com/rhinstaller/anaconda/pull/375

marmarek:
Rebased to F24.

marmarek:
Anything missing?

marmarek:
Why this was closed?

jcpunk:
This doesn't look like it got merged.

Any ideas why?

marmarek:
I'd like to know that too.
I can rebase it on the current version, if anyone can confirm there is a will to merge this feature.

Conan-Kudo:
I want to see this feature merged into pykickstart, because I want to merge [support for this into livecd-tools as well](https://github.com/livecd-tools/livecd-tools/pull/14).

@clumens Could you please consider allowing this feature in?

Conan-Kudo:
@dcantrell Could you please consider incorporating this into pykickstart?

Conan-Kudo:
@bcl Could you please re-review this and consider incorporating it into pykickstart?

jkonecny12:
@Conan-Kudo we just discussing how to deal this. The point is that Anaconda do not want to create an API for every thing which is part of the repo file. So we have a plan to instead be able to import repo files to the installation environment. Right now this is more an RFE than anything else. 
Hmm, I guess I should really create an RFE to get all the inputs.

jkonecny12:
Hello everyone, we think this issue could be solved by this proposal: https://hackmd.io/@prJIqOTeSxC-W0RNGKIGbQ/ByzIz3kIH/edit

Please look and comment there if you are interested.

PR #34, opened 2015-09-10
Empty instLangs

dashea:
For a kickstart with `%package --instLangs=` set packages.instLangs to "" instead of None.


vpodzime:
Looks good to me.

PR #36, opened 2015-09-21
Update doc urls

mmckinst:


mmckinst:
@bcl updated

vpodzime:
These all look good to me. I just think they could all be squashed into a single patch.

mmckinst:
@vpodzime squashed

PR #37, opened 2015-09-30
docs: Add `--activate` as an explicit network option to the list

cgwalters:
The docs cover `--activate`, but the option isn't included in the
table itself, so if someone visually scans for it they might miss it.

I think this is a bit clearer.

I was looking for `--activate` docs while trying to debug
an Anaconda issue.


clumens:
What's up with the .pot file change?

cgwalters:
Fixed.

PR #38, opened 2015-09-30
Allow GIDs to be specified in the user --groups list.

dashea:
Document the change to the usage of --groups in https://github.com/rhinstaller/anaconda/pull/379


vpodzime:
Looks good to me.

dashea:
Fixed a mistake I found (As for -> As of) and pushed.

PR #43, opened 2015-10-19
pykickstart: don't set bootProto if --noipv4

josefbacik:
If we specify a ipv6 only network configuration we still get bootproto=dhcp set,
which during kickstart makes networkmanager try to do a ipv4 dhcp which takes a
while to timeout.  We don't want bootproto=dhcp set when we specify --noipv4, so
make sure it gets zero'ed out when we see that command.

This is the V2 version that changes f24 and onwards.

### DIFF
--- a/pykickstart/commands/network.py
+++ b/pykickstart/commands/network.py
@@ -545,6 +545,19 @@ class F22_Network(F21_Network):
 
         return retval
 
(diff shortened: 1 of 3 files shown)


vpodzime:
Looks good to me.

PR #53, opened 2015-12-09
Fix class inheritance in RHEL6_VolGroup

cacoyle:
https://github.com/rhinstaller/pykickstart/blob/master/pykickstart/commands/volgroup.py#L174

`RHEL6_VolGroup` should inherit from `FC16_VolGroup` so that it understands disk reservation.


clumens:
Huh, looks like the RHEL6 patches didn't really get ported to master.

cacoyle:
Squashed commits down to a single one

clumens:
Merged, thanks.

PR #58, opened 2016-01-22
load.py initial test coverage

jikortus:
* The following test cases were created:
  - Load_To_String_TestCase - load content from file to string
  - Load_To_File_TestCase - load content from file to another file
  - Load_From_URL_To_Str_TestCase - load content from URL to string
  - Load_From_URL_To_File_TestCase - load content from URL to file

This is my first pull request + a few first unit tests, so any suggestions or ideas on how to improve those tests are very welcome.

### DIFF
--- a/pykickstart/load.py
+++ b/pykickstart/load.py
@@ -109,5 +109,5 @@ def _copy_file(filename, destination):
 
     try:
         shutil.copyfile(filename, destination)
-    except OSError as e:
+    except (OSError, IOError) as e:
         raise KickstartError(_('Error copying file: %s') % str(e))
--- a/tests/load.py
+++ b/tests/load.py
@@ -1,0 +1,120 @@
+import unittest
+import os
+import tempfile
+import six
+
+from pykickstart import load
+from pykickstart.errors import KickstartError
+from signal import SIGTERM
+
+class LoadTest(unittest.TestCase):
+    _content = """
+auth --enableshadow --passalgo=sha512
+graphical
+firstboot --enable
+ignoredisk --only-use=vda
+keyboard --vckeymap=cz --xlayouts='cz'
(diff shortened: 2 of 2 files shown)


clumens:
Thanks for the test case!  Can you please verify that it works with PYTHON=python2 as well?  That'll get run when the python2 version of the package gets made.  I'm seeing the following traceback when I do that:

======================================================================
ERROR: runTest (tests.load.Load_To_File_TestCase)
----------------------------------------------------------------------
  File "/home/clumens/src/pykickstart/tests/load.py", line 63, in runTest
    self.assertRaises(KickstartError, load.load_to_file, "/tmp/foo", "/tmp/bar")
  File "/usr/lib64/python2.7/unittest/case.py", line 511, in assertRaises
    callableObj(*args, **kwargs)
  File "pykickstart/load.py", line 71, in load_to_file
    _copy_file(location, destination)
  File "pykickstart/load.py", line 111, in _copy_file
    shutil.copyfile(filename, destination)
  File "/usr/lib64/python2.7/shutil.py", line 82, in copyfile
    with open(src, 'rb') as fsrc:
IOError: [Errno 2] No such file or directory: '/tmp/foo'

jikortus:
Thanks for pointing this out, I had not tried the tests explicitly with python2. The reason for the traceback was caused by the fact that IOException was raised instead of FileNotFoundError (which is derived from OSError caught even previously in _copy_file() in load.py) in python2 when a nonexistent file was opened. I fixed that in load.py (added IOExceptions to the exceptions that are caught when copying the file), I hope it is OK this way and doesn't break anything.

clumens:
Yeah that looks good.  Thanks!  I've pushed manually to avoid the merge commit.

jikortus:
Great, thank you!

PR #59, opened 2016-01-28
Do not log httpd messages in the load tests.

dashea:
By default, the python http server will print the httpd logs to stderr.
In our usage, those messages include inducing an error by feeding a
bunch of HTTPS data in as a HTTP request. This is ugly. Just skip the
logging entirely because who cares.


vpodzime:
Sounds and looks good to me.

PR #63, opened 2016-02-25
Rhel7 1240410 and 1311755

bcl:
This is a combined backport of the sshkey kickstart command and the sshpw --sshkey feature. The second depends on code from the 1st so may as well do both.


clumens:
Merged, thanks.  Would you be willing to also put together a PR for master so that branch knows that RHEL7 now supports this?

PR #70, opened 2016-03-12
Only show the deprecation warnings once.

bcl:
Keep track of the ones that have been shown and don't show them again.

### DIFF
--- a/pykickstart/base.py
+++ b/pykickstart/base.py
@@ -42,7 +42,7 @@ from pykickstart.i18n import _
 
 import six
 import warnings
-from pykickstart.errors import KickstartParseError, formatErrorMsg
+from pykickstart.errors import KickstartParseError, formatErrorMsg, deprecated
 from pykickstart.ko import KickstartObject
 from pykickstart.version import versionToString
 from pykickstart.parser import Packages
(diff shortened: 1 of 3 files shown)


clumens:
Doesn't this mean that fixing up the code using deprecated stuff will require tons of iterations?  That is, you'll get one warning about using something deprecated, fix that one, run your thing again to see where the next one is, etc.

dwlehman:
There's also something in the warnings module to handle issuing the warning only once IIRC.

bcl:
True, this would require you to search for other occurrences of what it is warning about, but I think that'd be nicer than repeating them.

Also, I didn't look too hard at the warnings module so I may have missed something.

clumens:
Is the following patch good enough for knocking down the pile of warnings when running anaconda?  It basically turns off all the pykickstart internal warnings so you should only get them for uses outside pykickstart.  If this isn't good enough, I've got another idea to try out.

diff --git a/pykickstart/__init__.py b/pykickstart/__init__.py
index aefb9e2..92a45c5 100644
--- a/pykickstart/__init__.py
+++ b/pykickstart/__init__.py
@@ -1,2 +1,6 @@
+# Ignore PendingDeprecationWarnings in the pykickstart module itself.  There's
+# nothing users of pykickstart can do about those.  Instead just print out ones
+# for users of the code.
 import warnings
-warnings.simplefilter("module", PendingDeprecationWarning)
+warnings.filterwarnings("once", category=PendingDeprecationWarning)
+warnings.filterwarnings("ignore", category=PendingDeprecationWarning, module="pykickstart")

clumens:
I've pushed the above patch.

PR #73, opened 2016-03-23
Kickstart mdraid chunk size support

vojtechtrefny:
Support for specifying chunk size for newly created mdraid arrays.


dwlehman:
Looks good to me, but should get a look from @clumens as well.

clumens:
This looks good for the master branch.  For the pykickstart-2 branch (which is where I am doing builds for Fedora from), there's going to need to be a minor change.  I think all that needs to be done is add_argument needs to be changed to add_option, and type=int needs to be changed to type="int".  If you submit that as a separate PR, I'll be happy to look it over and merge it.  Also you can run all the standard tests on that branch too.

PR #75, opened 2016-04-07
Enable coverage reporting for pykickstart tools

jikortus:
Changes in Makefile / setup.py / pykickstart.spec were made so as to enable pykickstart tools code coverage reporting. The tools were renamed to have .py extension (for the coverage), the extension will be renamed during RPM build so that the result after pykickstart installation remains unchanged.


pykickstart.spec
```diff
@@ -80,6 +80,14 @@ pushd %{py3dir}
...
 popd
 
+# pykickstart tools were renamed to have .py extension
+# (this is needed because of code coverage reporting)
+# -> must be renamed back for installation
+mv %{buildroot}/%{_bindir}/ksflatten.py %{buildroot}/%{_bindir}/ksflatten
+mv %{buildroot}/%{_bindir}/ksshell.py %{buildroot}/%{_bindir}/ksshell
+mv %{buildroot}/%{_bindir}/ksvalidator.py %{buildroot}/%{_bindir}/ksvalidator
+mv %{buildroot}/%{_bindir}/ksverdiff.py %{buildroot}/%{_bindir}/ksverdiff
+
```
clumens: Can you do this renaming in setup.py instead of in the spec file?

jikortus: I fixed that, I hope it is OK this way.

PR #78, opened 2016-05-02
Add support for --excludeWeakdeps option to %packages.

dashea:
This allows minimal installs, and reproducible ones, by not including
weak dependencies.

From https://bugzilla.redhat.com/show_bug.cgi?id=1331100

### DIFF
--- a/pykickstart/sections.py
+++ b/pykickstart/sections.py
@@ -34,7 +34,7 @@ from pykickstart.constants import KS_SCRIPT_PRE, KS_SCRIPT_POST, KS_SCRIPT_TRACE
                                   KS_MISSING_IGNORE, KS_MISSING_PROMPT
 from pykickstart.errors import KickstartParseError, formatErrorMsg
 from pykickstart.options import KSOptionParser
-from pykickstart.version import FC4, F7, F9, F18, F21, F22
+from pykickstart.version import FC4, F7, F9, F18, F21, F22, F24
 
 from pykickstart.i18n import _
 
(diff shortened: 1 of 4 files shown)


pykickstart/sections.py
```diff
@@ -254,6 +254,7 @@ def handleHeader(self, lineno, args):   # type: (PackageSection, int, List[str])
         op.add_argument("--default", dest="defaultPackages", action="store_true", default=False, introduced=F7)
         op.add_argument("--instLangs", default=None, introduced=F9)
         op.add_argument("--multilib", dest="multiLib", action="store_true", default=False, introduced=F18)
+        op.add_argument("--excludeWeakdeps", dest="excludeWeakdeps", action="store_true", default=False, introduced=F24)
```
dashea: The F24 constant needs to be imported in the "from pykickstart.version import ..." line near the top of the file. And I assume we'd want to go with F25, since 24 is just about done.

M4rtinK:
The new option should be also documented in the docs/kickstart-docs.rst file.

pykickstart/sections.py
dashea: So this is wanted for F24, I now know. Still need to import it though.

dashea:
Added the missing import

clumens:
Agreed with the documentation, and then this looks good to go.

dashea:
Added documentation here and in the other PR, and I think I spelled everything right this time

PR #79, opened 2016-05-02
Add support for --excludeWeakdeps option to %packages.

dashea:
This allows minimal installs, and reproducible ones, by not including
weak dependencies.

This is a pykickstart-2 port of https://github.com/rhinstaller/pykickstart/pull/78


dashea:
Fixed the obvious issues (missing import and bad test invocation) and re-pushed

PR #82, opened 2016-05-19
Adapt to the new version of mypy

dashea:
mypy's --use-python-path option has been removed, since the results were
often misleading. Move the existing type comments into .pyi stub files,
which are installed to /usr/lib/python*/site-packages, and run the type
checks based on those.

The test for this is probably going to fail since I haven't actually built this new mypy yet.


clumens:
Let me know when you have the new version built and I'll blast it out to the builders, and then we can get this retested.

clumens:
Jenkins, test this please

PR #85, opened 2016-05-25
Produce versioned docs, fixes #39

### LINKED ISSUE #39: RFE: Make docs more clear about when kickstart options were added
The current doc for kickstart options isn't very clearly versioned - there is just a list of options. It would be much appreciated if a note indicating when options and sub-options were added could be added to the docs. This would lead to being able to know which options work on which Fedora releases without having to dig through git history.

atodorov:
This is my initial take on this, to allow for easier documentation of commands and provide versioning.

I'm using Sphinx and it's autodoc extension to help me compile the text. In this scheme the commands description is implemented as a module doc text listing all possible options and values. Each class (e.g. RHEL6, RHEL7, Fedora20, Fedora 23, etc) is responsible for documenting its own syntax and acceptable values if there are any. 

I've made a proof of concept with authconfig and autopart commands. Let me know what you think before I continue documenting all of the rest (which is mostly a matter of copying the text that already exists).

For example the following will be added at the bottom of the autopart command description:
```
Syntax differences between versions:

class pykickstart.commands.autopart.F12_AutoPart(writePriority=100, *args, **kwargs)[source]

    autopart [--encrypted [--passphrase=<phrase>] [--escrowcert=<url> [--backuppassphrase=<passwd>]]]

class pykickstart.commands.autopart.F16_AutoPart(writePriority=100, *args, **kwargs)[source]

    autopart [--encrypted [--passphrase=<phrase>] [--escrowcert=<url> [--backuppassphrase=<passwd>]]] [--nolvm]

class pykickstart.commands.autopart.F17_AutoPart(writePriority=100, *args, **kwargs)[source]

    autopart [--encrypted [--passphrase=<phrase>] [--escrowcert=<url> [--backuppassphrase=<passwd>]]] [--nolvm] | [--type=<type>]

    where <type> is lvm, btrfs, plain

class pykickstart.commands.autopart.F18_AutoPart(writePriority=100, *args, **kwargs)[source]

    autopart [--encrypted [--passphrase=<phrase>] [--escrowcert=<url> [--backuppassphrase=<passwd>]] --cipher=<cipher>] [--nolvm] | [--type=<type>]

    where <type> is lvm, btrfs, plain

class pykickstart.commands.autopart.F20_AutoPart(writePriority=100, *args, **kwargs)[source]

    autopart [--encrypted [--passphrase=<phrase>] [--escrowcert=<url> [--backuppassphrase=<passwd>]] --cipher=<cipher>] [--nolvm] | [--type=<type>]

    where <type> is lvm, btrfs, plain, thinp

class pykickstart.commands.autopart.F21_AutoPart(writePriority=100, *args, **kwargs)[source]

    autopart [--encrypted [--passphrase=<phrase>] [--escrowcert=<url> [--backuppassphrase=<passwd>]] --cipher=<cipher>] [--nolvm] | [--type=<type>] [--fstype=<fstype>]

    where <type> is lvm, btrfs, plain, thinp, <fstype> is ext4, xfs, etc.

class pykickstart.commands.autopart.F9_AutoPart(writePriority=100, *args, **kwargs)[source]

    autopart [--encrypted [--passphrase=<phrase>]]

class pykickstart.commands.autopart.FC3_AutoPart(writePriority=100, *args, **kwargs)[source]

    autopart

class pykickstart.commands.autopart.RHEL6_AutoPart(writePriority=100, *args, **kwargs)[source]

    autopart [--encrypted [--passphrase=<phrase>] [--escrowcert=<url> [--backuppassphrase=<passwd>]] --cipher=<cipher>]

class pykickstart.commands.autopart.RHEL7_AutoPart(writePriority=100, *args, **kwargs)[source]

    autopart [--encrypted [--passphrase=<phrase>] [--escrowcert=<url> [--backuppassphrase=<passwd>]] --cipher=<cipher>] [--nolvm] | [--type=<type>] [--fstype=<fstype>]

    where <type> is lvm, btrfs, plain, thinp, <fstype> is ext4, xfs, etc.
```


docs/conf.py
```diff
@@ -0,0 +1,242 @@
...
+# General information about the project.
+project = u'Pykickstart'
+copyright = u'2016, Anaconda Team'
+
+# The version info for the project you're documenting, acts as replacement for
+# |version| and |release|, also used in various other places throughout the
+# built documents.
+#
+# The short X.Y version.
+version = '0.1'
```
bcl: You can hook this up to read the version from pykickstart.spec, eg. https://github.com/bcl/run-virt/blob/master/run-virt#L119

clumens:
I don't even know that it's necessary to have the syntax version summary.  There's the ksverdiff script that will generate that information for any two versions.  But if it is necessary, could it just provide the version and name for each class instead of listing the whole first line?  The way it is now looks more like API docs than end user docs.

docs/conf.py
atodorov: I was planning to fix this, submitted the PR as a POC to see what you guys think about the documentation format. However the link you've posted seems to be wrong.

atodorov:
Chris,
indeed it's API docs, I've just mixed it into the end-user docs. Kind of trying to satisfy both needs. I'll see if I can make it more user friendly.

Re ksverdiff - I don't think this is really popular, although it is very helpful. What I usually do is search for "anaconda kickstart syntax" or go directly to the kickstart-docs file and search what I need in it. And that doesn't give me any information about syntax differences between versions. The fact that we have issues open about this means others follow pretty much the same workflow.

atodorov:
@clumens - I don't see a way to instruct Sphinx not to print the "class module.Class" prefixes and I should probably drop these from the docs. However I want to keep the version differences available.

I was thinking about refactoring ksverdiff into methods and using them to generate diff between all commands defined in the module. I guess it is possible to iterate over all kickstart version constants and diff the respective command classes between each one, thus dynamically building the module docstring. How do you feel about this ?

@tflink - can you comment on this proposal as well (related to #39) ?

clumens:
I'd be fine with ksverdiff getting made into a library and a script that calls that library.  As far as it being not popular, that's too bad.  What syntax differences does it not report, and what issues are open about it?  I would like to attack those problems if so.

And then finally - are you talking about producing a document that lists the syntax differences from every version to every version?  So, 1 -> 2, 1 -> 3, 1 -> 4, 2 -> 3, 2 -> 4, 3 -> 4?  Seems like that could get pretty out of control.  There are a lot of versions.

atodorov:
@clumens I'd love for ksverdiff library to produce the full options (+ option values) listing and annotate sone of them with `New since XYZ` statements. If we can do this I think it will solve all problems.

clumens:
Is that something you'd like to work on, or perhaps I could do it if you outline what you'd like to see further.

clumens:
Regardless, I don't think that should block us from continuing on this versioned documentation thing.  I'm going to make a very picky comment or two on formatting for those and we can get this going.

tflink:
@atodorov it looks good to me from a user perspective, effectively solves the problem I first ran into so long as the docs are regularly updated (and I suspect they will be). Formatting isn't perfect but it definitely solves the problem and I can't really think of a better way to do this with sphinx :)

my only (nitpicky) suggestion would be to use something like "Syntax Versions" instead of "Syntax differences between versions" - differences between sounds like there is a diff instead of listing the different versions of commands.

pykickstart/commands/autopart.py
```diff
@@ -48,6 +110,9 @@ def parse(self, args):
         return self
 
 class F9_AutoPart(FC3_AutoPart):
+    """
+        ``autopart [--encrypted [--passphrase=<phrase>]]``
```
clumens: Does this mean that the docs for F9_AutoPart will only include the paragraphs for --encrypted and --passphrase= ?

pykickstart/commands/autopart.py
```diff
@@ -209,6 +288,13 @@ def _getParser(self):
         return op
 
 class F17_AutoPart(F16_AutoPart):
+    """
+        ``autopart [--encrypted [--passphrase=<phrase>]
+        [--escrowcert=<url> [--backuppassphrase=<passwd>]]]
+        [--nolvm] | [--type=<type>]``
+
+        where <type> is lvm, btrfs, plain
```
clumens: I'd like to see this look more like a sentence:

Where <type> is one of lvm, btrfs, or plain

pykickstart/commands/autopart.py
```diff
@@ -328,6 +430,15 @@ def parse(self, args):
...
 
 class F21_AutoPart(F20_AutoPart):
+    """
+        ``autopart [--encrypted [--passphrase=<phrase>]
+        [--escrowcert=<url> [--backuppassphrase=<passwd>]]
+        --cipher=<cipher>]
+        [--nolvm] | [--type=<type>] [--fstype=<fstype>]``
+
+        where <type> is lvm, btrfs, plain, thinp,
+        <fstype> is ext4, xfs, etc.
```
clumens: Likewise here, with some additional formatting to line up <fstype> under <type>.  Assuming that can be done.

clumens:
Here's another one to think about:  Right now the kickstart docs are just at some static URL that we can easily point people to.  After this change, kickstart docs will need to be generated.  Where do we store those?  Do we generate and store all versions somewhere?

pykickstart/commands/autopart.py
atodorov: Yes

atodorov:
> Here's another one to think about: Right now the kickstart docs are just at some static URL that we can easily point people to. After this change, kickstart docs will need to be generated. Where do we store those? Do we generate and store all versions somewhere?

Documenting all command versions with their respective syntax or using the "New since" annotations will introduce versioning right into the document text and remove the need for storing multiple documents. It will still be only 1 URL.

Migrating to Sphinx will likely mean the URL needs to change (GitHub pages or Read The Docs) but that should be fine.

atodorov:
> Is that something you'd like to work on, or perhaps I could do it if you outline what you'd like to see further.

@clumens I can work on this, just wanted to get a general direction of which is the preferred method to resolve the issue. ATM it appears to be user-friendly documentation (not looking like API docs), possibly generated with a modified version of `ksverdiff`. I'll see what I can do.

atodorov:
Hi everyone,
please take a look only at the last commit. Today I've tried a different approach so need to know what you think. 

I'm using the integrated help capabilities of ArgumentParser together with a custom formatter which produces pretty output for further processing with Sphinx. Try `python gendoc.py` and paste the output into `docs/index.rst` and then `make -C docs/ html`.

You will also notice how I mark options which changed behavior between versions in order to keep the whole history of changes.  

For the production version I'm planning to loop over all commands from the latest version and compile HTML from their help output. 

Comments and suggestions are welcome. I will continue working on this in the next few days if no objections.

clumens:
I definitely like where this is going.  I like that it puts the documentation right at the same point as the code.  I like that it means I can write a test case for requiring new options and commands to have documentation.  The key here is formatting the add_argument calls in such a way as to make sure the code continues to look pretty clean.  I think you're on the right track with that.

Could versions look more like "Fedora 20" instead of "Fedora20", or does it have to be a string without whitespace?

Also, I think after converting a couple of these we're going to start to see some common patterns emerge and add some helper functions for all this stuff.  But that's okay.

pykickstart/commands/autopart.py
```diff
@@ -76,9 +141,24 @@ def __str__(self):
...
-        op = KSOptionParser()
-        op.add_argument("--encrypted", action="store_true", default=False)
-        op.add_argument("--passphrase")
+        op = KSOptionParser(prog="autopart",  description="""
+Automatically create partitions -- a root (/) partition, a swap
+partition, and an appropriate boot partition for the architecture. On
+large enough drives, this will also create a /home partition.
+
+The ``autopart`` command can't be used with the logvol, part/partition, raid,
+reqpart, or volgroup in the same kickstart file.""")
```
bcl: Indenting this block would make it look nicer, and I think it'll strip the same # of spaces off when making the string.

bcl:
I agree, I like the direction this is going!

atodorov:
Here's an update of the progress so you can see where this is going. I'll look at formatting. 

Do we want all of the docs changes in one big commit or single commit per command ?

pykickstart/commands/autopart.py
atodorov: It's a multiline string and the spaces are transferred to the resulting rst text. I'll leave it as is for now and then look at it again to see how I can improve readability of source code comments.

clumens:
One big commit is probably too unwieldy, but one commit per command is probably too many commits.  I'd just do a couple at a time and send them off.

atodorov:
OK, let me finish the initial implementation and I will squash the commits when done.

clumens:
If you're locally committing one for each command, we might as well just push them that way.  No need to make extra work.

pykickstart/commands/bootloader.py
```diff
@@ -211,7 +205,8 @@ class F12_Bootloader(F8_Bootloader):
 
     def _getParser(self):
         op = F8_Bootloader._getParser(self)
-        op.add_argument("--lba32", dest="forceLBA", deprecated=True, action="store_true")
+        op.add_argument("--lba32", dest="forceLBA", action="store_true",
+                        deprecated=True, version="Fedora12")
```
clumens: I really don't like introducing these version= arguments (and the same sort of thing with remove_argument).  We already know which version added/removed/deprecated the argument - the version of the class the call was made from.  It seems likely there's some way we can automatically get this information.

atodorov: @clumens - I thought a bit about it but left it for later. At the moment I can easily perform some heuristics based on the class name, e.g.
```
version = self.__class__.__name__.replace('FC', 'F').replace('F', 'Fedora').split('_')[0]
```

Is this what you meant or there's more reliable versioning information which I'm not aware of at the moment ?

clumens: That is gross, too.  Okay I've looked around at the code a bit and I see we already treat deprecated= as both a bool and a version constant.  So let's:

(1) Change all the calls of deprecated=True to deprecated=<version constant>.  We'll still have to accept deprecated=True/False in options.py, and we probably can't do anything clever to figure out its help output, but I've supported it for so long it's gotta stay.
(2) Add a version=<version constant> argument to remove_argument.
(3) Somewhere add a conversion function that takes a version constant as an input and spits out a string suitable for feeding to help, and calling that in add_argument and remove_argument.  That way the help stuff doesn't leak out into all other functions.

Does this take care of everything?

pykickstart/commands/btrfs.py
```diff
@@ -126,14 +126,45 @@ def level_cb(value):
...
+subvolumes, should be the identifier of the subvolume's parent volume.
+
+``<mntpoint>``
+
+    Location where the file system is mounted.""")
+        op.add_argument("--noformat", dest="format", action="store_false",
+                        default=True, help="""
+    Use an existing BTRFS volume (or subvolume) and do not reformat the
+    filesystem.""")
+        op.add_argument("--useexisting", dest="preexist", action="store_true",
```
bcl: Not having this indented makes the source fairly ugly. It's simple to handle though:

`from textwrap import dedent`

and then wrap the multiline strings in a dedent() call. I'd indent them over to the level of the first arg of KSOptionParser, but not further.

pykickstart/commands/repo.py
```diff
@@ -170,44 +170,54 @@ def __str__(self):
...
+
+                            Note: If you want to enable one of the repos in
+                            /etc/anaconda.repos.d that is disabled by default
+                            (like "updates"), you should use --name= but none
+                            of the other options. anaconda will look for a repo
+                            by this name automatically. Providing a baseurl or
+                            mirrorlist URL will result in anaconda attempting
+                            to add another repo by the same name, which will
+                            cause a conflicting repo error.""")
         op.add_argument("--name", required=True, help="""
```
atodorov: @bcl - how does the new formatting look like ? I will format all new descriptions in this manner and then go back and update the others as well

clumens: This looks good to me.

bcl: Yeah, looks good, thanks!

pykickstart/options.py
```diff
@@ -126,11 +126,14 @@ def __init__(self, *args, **kwargs):
         """
         # Overridden to allow for the version kwargs, to skip help option generation,
         # and to resolve conflicts instead of override earlier options.
-        version = kwargs.pop("version", None)
+        version = kwargs.pop("version") # fail fast if no version specified
```
atodorov: @bcl, @clumens - I'm using this fail fast approach at a few places. Basically fail if no version/help parameters are provided. This will cause failures b/c there are commands and arguments for which I don't have documentation to copy.

atodorov:
@bcl, @clumens - I've updated formatting and versioning for all command and also adopted a fail fast approach - to traceback if some of the required parameters aren't provided. This however breaks the tests at the moment. I'll ping you on IRC and try to provide reasonable defaults for the missing help options so that the tests pass again. Afterwards we can merge this if no objections. 

Compiling the docs and further sanity testing for all command classes and their help documentation can be in a separate PR b/c this one is already too big. Please review and comment.

clumens:
Yes, this PR has gotten too big to do anything with.  How are tests broken?  If we can do something quickly to get them working again (or even just ignored for the moment), I'd be fine with doing that, merging this PR, and then dealing with the test cases.

clumens:
Jenkins, test this please.

clumens:
Okay, I had jenkins run the tests to see what's busted.  I see some undefined variables that should definitely get fixed up and squashed, but the rest of it could be handled in a followup PR (or I'd just do it myself).  I don't see anything big holding this one up.

atodorov:
@clumens - I'm adding the final touches to fix the tests. Only 1 failure left.

atodorov:
Jenkins, test this please.

atodorov:
@clumens, @bcl - I have the tests passing locally. Please review. In particular check more thoroughly volgroup.py, raid.py, timezone.py and sshpw.py - look for my comments in the code.

Most of the changes are related to adding new help description and versions. In a few places I've also added positional arguments to the option parser. This helps with documentation and also gives a better structure to all the parsed arguments. Let's clean this up if necessary and merge before I go any further with the rest of the documentation.

clumens:
I definitely remember looking at adding the positional arguments when I moved to argparse, then backing that out for some reason.  I do not recall what the reason was.  It may have had something to do with anaconda.  Regardless, if it's working for you with the tests then it's probably alright.  I'll comment on the code comments separately.  Then I really hope we can get this thing merged.  Geez.

pykickstart/commands/timezone.py
```diff
@@ -141,11 +175,8 @@ def __init__(self, *args, **kwargs):
         F18_Timezone.__init__(self, *args, **kwargs)
         self.ntpservers = kwargs.get("ntpservers", list())
 
-    def _getParser(self):
-        op = FC6_Timezone._getParser(self)
-        op.add_argument("--nontp", action="store_true", default=False)
-        op.add_argument("--ntpservers", type=commaSplit)
-        return op
+#todo: I've removed _getParser() but this method also looks like above
+# something to do with lists and sets
```
clumens: This is fine - I don't see any reason why there's the duplicate _getParser method.  The distinct parse methods need to stay, though.  We made a change where you can have the same NTP server listed multiple times for some reason, whereas other releases do not allow for that.

pykickstart/commands/sshpw.py
```diff
@@ -148,5 +184,11 @@ class F24_SshPw(F13_SshPw):
 
     def _getParser(self):
         op = F13_SshPw._getParser(self)
-        op.add_argument("--sshkey", action="store_true", default=False)
+#todo: this description doesn't make sence. Is this the same as the sshkey command
+# or different. Why is it a boolean flag instead of accepting the key content ?
```
clumens: The sshpw command only affects the installation environment - it defines a user and ssh key that can be used while anaconda is running.  It has no effect on the finished system.  The sshkey command allows you to associate an ssh key with a user on the finished system.  That user could either already exist or be one created with the user command.

And then if you give --sshkey, the password you provided is interpreted to be the key.  So, "sshpw --username=A blah" means user A has the password of blah.  "sshpw --username=A --sshkey blah" means user A has an ssh key of blah, and no password.

pykickstart/commands/raid.py
```diff
@@ -414,16 +499,46 @@ def __init__(self, writePriority=131, *args, **kwargs):
...
+                        default=False, help="""
+                        Specify that this RAID device should be encrypted.""")
+        op.add_argument("--passphrase", version=F9, help="""
+                        Specify the passphrase to use when encrypting this RAID
+                        device. Without the above --encrypted option, this option
+                        does nothing. If no passphrase is specified, the default
+                        system-wide one is used, or the installer will stop and
+                        prompt if there is no default.""")
+#todo: it looks like this needs to be removed earlier, see F9_RaidData
+# instead of F14. Test fails otherwise. Why doesn't it fail in Jenkins ?
```
clumens: I don't know why they don't fail, but the argument can't be removed earlier.  It was still recognized by the pykickstart shipped with anaconda all the way up to F14, even if it didn't do anything.

atodorov: OK, I've removed this (and related comments). Can you kick off a Jenkins build (b/c I can't) and see if this passes. I will later investigate why it fails for me locally.

clumens:
Jenkins, test this please.

clumens:
Jenkins, test this please.

atodorov:
I've updated the code with pylint fixes, everything should pass now.

clumens:
Jenkins, test this please.

pykickstart/commands/raid.py
```diff
@@ -414,16 +499,43 @@ def __init__(self, writePriority=131, *args, **kwargs):
...
+                break
+        return op
+
 class F9_Raid(F7_Raid):
     removedKeywords = F7_Raid.removedKeywords
     removedAttrs = F7_Raid.removedAttrs
 
     def _getParser(self):
         op = F7_Raid._getParser(self)
-        op.add_argument("--bytes-per-inode", deprecated=True)
```
clumens: Is this line a mistake?  See the results from the most recent jenkins run.

atodorov: Yes, that's why these tests were failing earlier for me. Thanks for catching it.

clumens:
Jenkins, test this please.

PR #87, opened 2016-06-13
[rhel7-branch] Support timezone command usage without timezone specification

M4rtinK:
Make it possible to use the timezone command without a timezone
specification.

This way the timezone command can be used with the --utc flag
to set hardware clock time format while still requiring the user
to select a timezone manually in the Anaconda UI.

Related: rhbz#1312135

### DIFF
--- a/pykickstart/commands/timezone.py
+++ b/pykickstart/commands/timezone.py
@@ -53,6 +53,8 @@ class FC3_Timezone(KickstartCommand):
         op.add_option("--utc", dest="isUtc", action="store_true", default=False)
         return op
 
+    # Caution: This method is overridden in the RHEL7_Timezone class
+    # by a new implementation not calling this method.
     def parse(self, args):
         (opts, extra) = self.op.parse_args(args=args, lineno=self.lineno)
         self._setToSelf(self.op, opts)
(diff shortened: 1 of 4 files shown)


M4rtinK:
Hope this is not too crazy.

Also the accompanying Anaconda PR is https://github.com/rhinstaller/anaconda/pull/670.

pykickstart/commands/timezone.py
```diff
@@ -35,6 +35,12 @@ def __init__(self, writePriority=0, *args, **kwargs):
         self.isUtc = kwargs.get("isUtc", False)
         self.timezone = kwargs.get("timezone", "")
 
+        # tracks if the timezone command requires a timezone specififcation
+        # - this has been the case up to the F18 version of the command
+        # - this variable has been added to make it possible for later versions
+        #   of the command to drop this requirement
+        self._timezone_spec_required = True
```
vpodzime: This would make more sense to me as a class attribute. But in general, I don't think these changes are worth reusing 3 lines of code in the ``parse()`` method. The RHEL7 version of the command could just have its own ``parse()`` method defined instead.

clumens:
In general, I prefer using the inheritance tree instead of variables for specifying whether things are supported in a particular version.  Sometimes that involves being a little wordier, but oh well.  For the RHEL7 version of this patch, I think the RHEL7 class's parse method should handle the possibility of not giving a timezone.  And then for the master version, you'll need to have the same thing in the RHEL7 class and whichever class will be the first in Fedora to support this.

M4rtinK:
>In general, I prefer using the inheritance tree instead of variables for specifying whether things are supported in a particular version. Sometimes that involves being a little wordier, but oh well. For the RHEL7 version of this patch, I think the RHEL7 class's parse method should handle the possibility of not giving a timezone. 

Sure - its just that in this case we need to prevent the FC3 parse() method from being called, otherwise we would always get an exception regardless to what the parse() method in the RHEL7 variant does.

So I've changed the PR to implement Vratislavs suggestion - I've just implemented all the stuff the "older" classes do in RHEL7 command class parse() method the way I need it - without calling the ancestors method. 

I've also added a note to the possibly skipped parse() methods so that people don't wonder why they are not being called in RHEL 7.3+.

>And then for the master version, you'll need to have the same thing in the RHEL7 class and whichever class will be the first in Fedora to support this.

Thanks! I wondered how this should be done for the master version - especially with the RHEL7 command variant. I guess it makes sense due to ksvalidator (it needs to be able to validate RHEL7 kickstarts).

clumens:
I think this is fine for RHEL7 - having the _getParser method is supposed to help somewhat with code duplication, and if the parse methods are too complicated, we should work to reduce that.  I think the current patch looks good.

clumens:
And then for master, you basically have to do two things:  (1) Add this patch to show that RHEL7 now supports it, and (2) Do something very similar adding a new F24 (F25?) class.  But yeah, especially for F24 the parse method is starting to get complex.  At the least, this establishes a point in time where we can start inheriting again.  It's not going to get more complicated.

M4rtinK:
I've added a check that throws KickstartParseError when only "timezone" is used in a kickstart & noted this in the documentation.

I've also added unit tests for the new behavior and (hopefully) improved the error messages a bit.

M4rtinK:
Pushed; thanks for the review & feedback!

PR #88, opened 2016-06-29
Enable Travis-CI

atodorov:
The only drawback I'm currently seeing is that multiprocessing and coverage don't play well in Travis. I frequently get some JSON errors or timeouts, see https://travis-ci.org/atodorov/pykickstart/builds/141046781.

However I've seen the same thing locally and I also think there's something wrong with the coverage generation from nose. The tools tests for example execute a subprocess with "coverage run ..." instead of relying on nose. I'll take a look at this in another PR. Maybe we'll want to disable multiprocessing for now, what do you think ?

### DIFF
--- a/tests/pylint/runpylint.py
+++ b/tests/pylint/runpylint.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python3
+#!/usr/bin/env python3
 
 import sys
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,16 @@
+language: python
+python:
+  - "3.5"
+
+install:
+  - pip install --upgrade pip setuptools
+  - pip install coverage mypy-lang nose ordered-set polib pylint requests six
+  - pip install https://github.com/rhinstaller/pocketlint/zipball/master
+
+script:
+  - make ci
+
+notifications:
+  email:
+    on_success: change
+    on_failure: change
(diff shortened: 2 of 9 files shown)


tests/pylint/runpylint.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/python3
+#!/usr/bin/env python
```
bcl: You can't do this. 'python' should always be python2.

atodorov: Well that contradicts the code b/c without my change it is currently python3.

bcl: Contradicts what code? What environment are you running this in? Anything that changes python to python3 is going to cause no end of problems. Use python3 for python3 code.

atodorov: Hi Brian,
see your previous comment. Did you say 'python2' by mistake ? Anyway, I'm running this inside a virtualenv (both locally and in Travis CI) with Python 3.5. However the interpreter isn't always called 'python3' hence the change.
The only thing I can think off is using
```
if not six.PY3:
    sys.exit(1)
```

at the top of this script. How about the others ?

atodorov: Forget my previous comment, I'm using `/usr/bin/env python3` now.

atodorov:
This is now ready to merge if you approve all changes.

PR #89, opened 2016-06-29
Fix some code smells

atodorov:
I've found landscape.io which appears to be a nice tool. This PR fixes some of the issues it found. The full report is available at https://landscape.io/github/atodorov/pykickstart. Some of the issues will be fixed once #85 is merged.


pykickstart/commands/zerombr.py
```diff
@@ -46,7 +46,7 @@ def _getParser(self):
         return op
 
     def parse(self, args):
-        (_ns, extra) = self.op.parse_known_args(args=args, lineno=self.lineno)
+        (_, extra) = self.op.parse_known_args(args=args, lineno=self.lineno)
```
clumens: This will conflict with the import of the translation function _.  I'd prefer to just say:

extra = self.op.parse_known_args(args=args, lineno=self.lineno)[1]

translation-canary/translation_canary/translated/__init__.py
```diff
@@ -146,7 +146,7 @@ def testSourceTree(srcdir, releaseMode=False, modifyLinguas=True):
     success = True
     srcdir = os.path.normpath(srcdir)
 
-    for dirpath, _dirnames, paths in os.walk(srcdir):
+    for dirpath, _, paths in os.walk(srcdir):
```
clumens: Our pylint config is set up to recognize that anything starting with an underscore is an unused variable.  I'd prefer to not have to go through and change all these.

atodorov:
@clumens - can you also take a look at the latest report here:
https://landscape.io/github/atodorov/pykickstart/30/messages

in particular the Unused import, Undefined name and Cyclic import message types. Is there anything else we could improve ?

clumens:
NAK to shortening long lines.  It's 2016 and I've got very wide monitors, with very big vim session in them, and I'm not going to deal with convoluted indentation schemes for the dubious benefit of being confined into very narrow terminals.

clumens:
Unused import - the two __init__.py ones are false positives.  Those are there so you can get everything just by importing "pykickstart.commands", say.  The others are probably legitimate.  Try removing them and seeing if mypy still passes.  I think those are from before we had the separate type annotation files.

Undefined name - I think this is a false positive, too.  That code will only run on python2, where unicode is defined.  If there's a better way to do this I would be happy to hear about it, because pylint complains about the same thing.

Cyclic imports - Definitely valid, don't know if it's a potential bug or not.  I also have no clever ideas about how that could be fixed.  Could the cycles be broken by putting a guard on some import?

atodorov:
> Cyclic imports - Definitely valid, don't know if it's a potential bug or not. I also have no clever ideas about how that could be fixed. Could the cycles be broken by putting a guard on some import?

I think this is the only thing left to worry about at the moment. I'll see what can be done to minimize this but that's probably a separate pull request. Everything else should be good to go as is.

atodorov:
@clumens please see the last 2 commits. I've managed to resolve all cyclic imports. 

In the first commit `pykickstart.handlers` was used only to get its path so I've replaced this by constructing the path based on the file name of the caller. 

In the second commit `pykickstart.handlers.control` uses the information from `__init__.py` to construct the commandMap and dataMap objects. In this case I'm listing the directory and loading the module objects based on their names.

In fact I can drop `9cf649b` entirely and everything will still be fine b/c `pykickstart.handlers.__init__.py` doesn't import anything anymore. Do you want me to do this or prefer not loading the module just to get its path ?

All tests PASS locally and landscape.io report looks much better now:
https://landscape.io/github/atodorov/pykickstart/43/messages

clumens:
I'm fine with the patches as they are.  Merging.

PR #90, opened 2016-06-30
Refactor ksvalidator and its tests

atodorov:
I've cleaned up a bit and converted ksvalidator to behave more like a module, not a script so we can test it more easily.


atodorov:
CC @jikortus , FYI

tests/tools/ksvalidator.py
```diff
@@ -1,15 +1,29 @@
...
 import os
+import tempfile
+from unittest import TestCase
+import unittest.mock as mock
+from tools import ksvalidator
+from tests.tools.utils import mktempfile
+from pykickstart.version import versionMap
 
-class No_Parameters_TestCase(KsvalidatorTest):
+class RemoveNonExistingFileTestCase(TestCase):
```
bcl: Should probably follow the existing style and keep using underscores.

bcl:
I like the refactoring for easier testing, but you dropped the coverage support. Was that on purpose?

atodorov:
@bcl - coverage is collected by the calling process, in this case `python -m nose --with-coverage` and it's not lost. It was wrong to run under coverage and also start the coverage process from inside the test, I've simplified things.

FYI, if we really need to collect coverage from subprocesses then we need the `COVERAGE_PROCESS_START` variable as it is done in the anaconda test suite. For more info see
https://coverage.readthedocs.io/en/coverage-4.1/subprocess.html

tools/ksvalidator.py
```diff
@@ -43,68 +43,81 @@ def cleanup(dest, fn=None, exitval=1):
...
-except KickstartError:
-    print(_("General kickstart error in input file"))
-    cleanup(destdir, processedFile)
-except Exception as e:
-    print(_("General error in input file:  %s") % e)
-    cleanup(destdir, processedFile)
+    return exitval
+
+def main(argv=sys.argv[1:]):
+    op = argparse.ArgumentParser(usage="%(prog)s [options] ksfile", add_help=False)
```
atodorov: Sorry but I don't really get you comment. Do you mean to move the argument parser out of the main function so it can be documented automatically ? If this is the case I'll look into it separately b/c I have to continue my work on the kickstart commands documentation anyway.

For this PR I think it can stay as is.

PR #93, opened 2016-07-14
Master version of the timezone-without-arguments patch

M4rtinK:
This adds a separate RHEL7 & F25 version of the modified timezone command & the corresponding tests, as suggested in:

https://github.com/rhinstaller/pykickstart/pull/87#issuecomment-225897780

### DIFF
--- a/pykickstart/commands/timezone.py
+++ b/pykickstart/commands/timezone.py
@@ -69,6 +69,8 @@ class FC3_Timezone(KickstartCommand):
                         Timezone name, e.g. Europe/Sofia.""")
         return op
 
+    # Caution: This method is overridden in the RHEL7_Timezone & F25_Timezone classes
+    # by a new implementation not calling this method.
     def parse(self, args):
         (ns, extra) = self.op.parse_known_args(args=args, lineno=self.lineno)
         self.set_to_self(ns)
@@ -157,6 +159,8 @@ class F18_Timezone(FC6_Timezone):
                         """)
         return op
 
+    # Caution: The parse() method is overridden in the RHEL7_Timezone & F25_Timezone classes
+    # by a new implementation not calling this method.
     def parse(self, args):
         FC6_Timezone.parse(self, args)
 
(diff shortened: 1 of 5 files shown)


pykickstart/commands/timezone.py
```diff
@@ -182,3 +188,92 @@ def parse(self, args):
...
+# classes and avoid calling the ancestors parse(), as it otherwise done in all
+# other command child classes.
+
+class RHEL7_Timezone(F18_Timezone):
+    def __init__(self, writePriority=0, *args, **kwargs):
+        F18_Timezone.__init__(self, writePriority, *args, **kwargs)
+
+    def parse(self, args):
+        (opts, extra) = self.op.parse_args(args=args, lineno=self.lineno)
+        self._setToSelf(self.op, opts)
```
clumens: These two lines should be:

ns = self.op.parse_args(args=args, lineno=self.lineno)
self.set_to_self(ns)

pykickstart/commands/timezone.py
```diff
@@ -182,3 +188,92 @@ def parse(self, args):
...
+
+        return self
+
+class F25_Timezone(F18_Timezone):
+    def __init__(self, writePriority=0, *args, **kwargs):
+        F18_Timezone.__init__(self, writePriority, *args, **kwargs)
+
+    def parse(self, args):
+        (opts, extra) = self.op.parse_args(args=args, lineno=self.lineno)
+        self._setToSelf(self.op, opts)
```
clumens: Same here.

pykickstart/commands/timezone.py
```diff
@@ -182,3 +188,92 @@ def parse(self, args):
...
+        # -> due to the FC3 parse() method that would be eventually called,
+        #    which throws an exception if no timezone specification is provided
+        # - we implement the relevant functionality of the ancestor methods here
+
+        if len(extra) > 1:
+            error_message = _("One or zero arguments are expected for the %s command") % "timezone"
+            raise KickstartParseError(formatErrorMsg(self.lineno, msg=error_message))
+
+        if len(extra) > 0:
+            self.timezone = extra[0]
```
clumens: And then instead of checking extra here, you need to look for ns.timezone.

pykickstart/commands/timezone.py
```diff
@@ -182,3 +188,92 @@ def parse(self, args):
...
```
clumens: Same here.

M4rtinK:
Updated the PR with the ns/extra changes. :)

clumens:
It looks like you also need to override _getParser at some level to make the timezone argument optional.

clumens:
Note: the reason this check is failing is because of a bug in the patch, and that's because master has all that docs stuff added.  This resulted in some changes to add_argument calls.  In particular, timezone is now grabbed via an add_argument call instead of looking at extra.

M4rtinK:
Pushed a new version:
- reimplemented _getPArser in the F25 & RHEL7 command versions
- the F25 version now subclasses the F23 command while the RHEL7 version still subclasses the F18 command version, as in the RHEL 7 patch version

BTW, what about tests ? Does the test case class inheritance cover this (class F25_TestCase(F23_TestCase, etc.)) or should I also "duplicate" the unit-testing code to make sure the "new" parser is adequately covered ?

M4rtinK:
So **now** the new version of the PR should be actually pushed. ;-)

PR #95, opened 2016-07-15
A bunch of static checker suggested improvements

M4rtinK:
As there are already some substantial changes in the Pykickstart 3 code base then it might make sense to also possibly also improve the code formatting when we are at it. :)

So I've run a battery of static checkers (Pylint, pyflakes, pep8, etc.) on the master branch and implemented the suggestions that looked sensible.

If some of the changes *don't* look sensible, let me know and I'll remove them from the PR. :)

### DIFF
--- a/pykickstart/commands/liveimg.py
+++ b/pykickstart/commands/liveimg.py
@@ -73,8 +73,7 @@ class F19_Liveimg(KickstartCommand):
             structured) the first *img file inside LiveOS will be mounted and
             used to install the target system. The URL may also point to a
             tarfile of the root filesystem. The file must end in .tar, .tbz,
-            .tgz, .txz, .tar.bz2, tar.gz, tar.xz""",
-            version=F19)
+            .tgz, .txz, .tar.bz2, tar.gz, tar.xz""", version=F19)
         op.add_argument("--url", metavar="<url>", required=True, version=F19,
                         help="""
                         The URL to install from. http, https, ftp and file are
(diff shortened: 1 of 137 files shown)


pykickstart/commands/liveimg.py
```diff
@@ -39,8 +39,8 @@ def __eq__(self, other):
             return False
 
         return self.url == other.url and self.proxy == other.proxy and \
-               self.noverifyssl == other.noverifyssl and \
-               self.checksum == other.checksum
+            self.noverifyssl == other.noverifyssl and \
+            self.checksum == other.checksum
```
clumens: I like all the conditions lined up to the right of the return here.  That occurs a handful of other places in this mega patch, too.  Aside from that everything looks great to me.

jkonecny12:
Looks great to me too. Good job!

pykickstart/commands/liveimg.py
M4rtinK: Should be changed now in all the places where it occurs. :)

atodorov:
@M4rtinK - can you add your tools to `Makefile` and `.travis.yml` so they get run in CI as well ?

M4rtinK:
@atodorov Well, its currently a bunch of vim plugins that that I use, so it is not really directly applicable. 

But I'm sure the plugins are just thin wrappers for the respective static checking utilities, which could then be called from the `Makefile` & CI. So I'll take a look what am I actually using. :)

PR #96, opened 2016-07-21
Mutation testing improvements

atodorov:
I'm experimenting with mutation testing, which in short, modifies the code under test (e.g. True becomes False, etc) and expects the test suite to fail. This means the test suite is good enough to catch the mutation and kill the mutant. In other words if we accidentally introduced a bug there our test suite will detect it. 

The first few commits are cosmetics, then I start adding new tests and modifying the code to kill remaining mutants. 

NB: the tool I'm using is called Cosmic Ray and is still under heavy development. I'm also executing it against the smallest possible chunks b/c there are around 5,500 possible mutations against the entire pykickstart codebase. My estimate is the entire test run will take around a week.

### DIFF
--- a/pykickstart/base.py
+++ b/pykickstart/base.py
@@ -293,7 +293,7 @@ class BaseHandler(KickstartObject):
         """Return a string formatted for output to a kickstart file."""
         retval = ""
 
-        if self.platform != "":
+        if self.platform:
             retval += "#platform=%s\n" % self.platform
 
         retval += "#version=%s\n" % versionToString(self.version)
--- a/pykickstart/commands/authconfig.py
+++ b/pykickstart/commands/authconfig.py
@@ -26,7 +26,7 @@ class FC3_Authconfig(KickstartCommand):
     removedAttrs = KickstartCommand.removedAttrs
 
     def __init__(self, writePriority=0, *args, **kwargs):
-        KickstartCommand.__init__(self, *args, **kwargs)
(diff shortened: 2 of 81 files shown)


atodorov:
@clumens @bcl @vpodzime ping. Can you review this PR b/c it's becoming a bit large and I have quite a few more tests to add. Basically I'm going over each source file and trying to add as many tests as I can in order to not leave any mutated versions around. Also I'm trying to boost code coverage to 100% where it's less.

clumens:
The str != "" patch has me a little concerned.  Isn't this going to cause problems where sometimes a string could be "" and sometimes it could be None?  I believe we've got places where we make a distinction between the two, if not in pykickstart then in anaconda.

clumens:
I guess I don't really have anything against this PR.  I find some of the code reorganizations a little pointless and pedantic, but don't really care one way or another.  Increasing test coverage is worth it.

atodorov:
> The str != "" patch has me a little concerned. Isn't this going to cause problems where sometimes a string could be "" and sometimes it could be None? I believe we've got places where we make a distinction between the two, if not in pykickstart then in anaconda.

I've seen a few places where stuff can be both None or 0 for example and there's a distinction between them but I didn't see a place where this applies for the strings I've changed. Maybe I've missed something but there's no way in telling since all tests PASS. Also note that when a str is None then 
str != "" evaluates to True and executes the same piece of code as if the string had a real value. I believe this is what we want anyway.

atodorov:
@clumens ping, merge ?

PR #106, opened 2016-10-06
Fix Python 2 builds by assigning to KSOptionParser.version properly

atodorov:
the underlying problem is in ArgumentParser.__init_ and its
deprecated version argument which was resetting self.version to
None.

@clumens this makes all tests run again under both Python 3 and Python 2.


InStallker:

InStallker:

clumens:
Outstanding, thanks.

PR #108, opened 2016-10-09
Disable HelpAndDescription_TestCase

atodorov:
@clumens I've temporarily disabled this test case. I will also go ahead and file issues for each individual module so we don't forget to finally update the descriptions.

Note: on my system I wasn't able to reproduce the failures. I think this has to do with how mock finds the imported modules where it needs to apply the patch. With the current changes I'm able to generate the list of errors as expected. Let me know how that works for you.


InStallker:

InStallker:

PR #111, opened 2016-10-17
Py2 add noninteractive

jkonecny12:
To prevent kickstart tests timeouts I need to add non-interactive mode which will kill installation instead of waiting on user interaction.

Anaconda patch will follow after the `anaconda.py` PR will be merged.

### DIFF
--- a/docs/kickstart-docs.rst
+++ b/docs/kickstart-docs.rst
@@ -593,6 +593,15 @@ graphical
 Perform the kickstart installation in graphical mode. This is the
 default.
 
+``graphical [--non-interactive]``
+
+``--non-interactive``
+
+    Perform the installation in a completely non-interactive mode.
+    This mode will kill the installation when user interaction will be
+    required. Behavior of this mode is similar to ``cmdline`` or
+    ``text --non-interactive`` modes but with graphics.
+
 
 halt
 ----
@@ -1973,6 +1982,13 @@ text
 
 Perform the kickstart installation in text mode. Kickstart installations
 are performed in graphical mode by default.
+
+``text [--non-interactive]``
+
+``--non-interactive``
+
+    If present, run the installation in completely non interactive text
+    mode. This mode behaves same as the ``cmdline`` mode.
 
 
 timezone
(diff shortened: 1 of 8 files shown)


jkonecny12:
I will add tests but it looks like there is a bug somewhere. I'll look on that tomorrow.

docs/kickstart-docs.rst
```diff
@@ -593,6 +593,15 @@ graphical
 Perform the kickstart installation in graphical mode. This is the
```
vpodzime: Cannot add a comment for the commit message so I'm adding it here. First of all, it's an option not a param(eter). And we don't have graphical modes, we only have one graphical mode and a text mode. Sorry for nitpicking here, but I really find the commit message confusing.

pykickstart/commands/displaymode.py
```diff
@@ -56,7 +57,54 @@ def parse(self, args):
...
+        (opts, extra) = self.op.parse_args(args=args, lineno=self.lineno)
+        self.set_to_self(self.op, opts)
+
+        if len(extra) > 0:
+            msg = _("Kickstart command %s does not take any arguments") % self.currentCmd
+            raise KickstartParseError(formatErrorMsg(self.lineno, msg))
+
+        if self.currentCmd == "cmdline" and self.nonInteractive:
+            msg = _("Kickstart command cmdline does not support --non-interactive parameter")
+            raise KickstartParseError(formatErrorMsg(self.lineno, msg))
```
vpodzime: I believe this should call the parent class' ``parse()`` method and just perform the extra checks.

docs/kickstart-docs.rst
jkonecny12: Sorry for that I wrote it in rush and I did mistakes. I hope it's good now.

jkonecny12:
Updated

pykickstart/commands/displaymode.py
```diff
@@ -66,3 +69,37 @@ def parse(self, args):
...
+        FC3_DisplayMode.__init__(self, writePriority, args, kwargs)
+        self.op = self._getParser()
+        self.displayMode = kwargs.get("displayMode", None)
+        self.nonInteractive = kwargs.get("nonInteractive", False)
+
+    def __str__(self):
+        retval = super(F26_DisplayMode, self).__str__()
+
+        if self.nonInteractive:
+            retval = retval[:-1]
```
clumens: Using strip here would make me feel better.

jkonecny12: Nice solution, thank you. I'll change that now.

jkonecny12:
Updated PR. Changed to `rstrip()`

PR #113, opened 2016-10-19
More test coverage

### LINKED ISSUE #112: Error parsing timezone command for RHEL7
Source kickstart:
`timezone --utc America/Los_Angeles`

Error:
```
The following problem occurred on line 18 of the kickstart file:

One or zero arguments are expected for the timezone command
```

I think the problem is here:
```
        ...
        if ns.timezone:
            if len(ns.timezone) == 1:
                self.timezone = ns.timezone[0]
            elif len(ns.timezone) > 1:
        ...
```
ns.timezone holds name of the timezone, so in my case len(ns.timezone) was 19.
I would remove the whole `if ns.timezone:` case, but not 100% sure.

atodorov:
Additional test coverage and refactoring. Also fixes #112

### DIFF
--- a/pykickstart/commands/autopart.py
+++ b/pykickstart/commands/autopart.py
@@ -206,8 +206,6 @@ class RHEL6_AutoPart(F12_AutoPart):
             conflicting_command = "volgroup"
         elif self.handler.logvol.seen:
             conflicting_command = "logvol"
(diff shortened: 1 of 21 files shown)


InStallker:

InStallker:

PR #119, opened 2016-10-26
bugfix: Here should be only one positional argument - one password

Frodox:


clumens:
I don't see any test cases that are failing because of this bug.  Am I just missing one?  If not, we should really add a test case for this - it's come up at least one other time now.

atodorov:
@clumens yes you are missing.

```
ERROR: runTest (tests.commands.rootpw.F18_TestCase)
----------------------------------------------------------------------
  File "/home/atodorov/repos/git/pykickstart/tests/commands/rootpw.py", line 75, in runTest
    self.assert_parse("rootpw --lock", "rootpw --lock\n")
  File "/home/atodorov/repos/git/pykickstart/tests/baseclass.py", line 166, in assert_parse
    obj = parser.parse(args[1:])
  File "/home/atodorov/repos/git/pykickstart/pykickstart/commands/rootpw.py", line 130, in parse
    (ns, extra) = self.op.parse_known_args(args=args, lineno=self.lineno)
  File "/home/atodorov/repos/git/pykickstart/pykickstart/options.py", line 276, in parse_known_args
    return ArgumentParser.parse_known_args(self, *args, **kwargs)
  File "/opt/rh/rh-python35/root/usr/lib64/python3.5/argparse.py", line 1758, in parse_known_args
    namespace, args = self._parse_known_args(args, namespace)
  File "/opt/rh/rh-python35/root/usr/lib64/python3.5/argparse.py", line 1993, in _parse_known_args
    ', '.join(required_actions))
  File "/home/atodorov/repos/git/pykickstart/pykickstart/options.py", line 257, in error
    raise KickstartParseError(formatErrorMsg(self.lineno, msg=message))
pykickstart.errors.KickstartParseError: The following problem occurred on line 0 of the kickstart file:

the following arguments are required: <password>
```

the password argument is not required for F18 when `--lock` is specified. I'm not sure about sshpw at the moment though.

A side note: I can enable Travis-CI so we can get tests executed for external contributions as well.

atodorov:
> runTest (tests.commands.rootpw.F18_TestCase) ... ok
> 
> So, do I have the wrong version of something installed?

Can you tell us about your environment ? I'm using Python 3.5 with the latest master and I get failures for both rootpw and sshpw if I apply the changes from this PR.

Frodox:
> I get failures for both rootpw and sshpw if I apply the changes from this PR.

Ok, but [currently](http://pykickstart.readthedocs.io/en/latest/commands_rhel.html?highlight=rootpw#rootpw) it looks like 

`rootpw [--iscrypted] ... [<password> [<password> ...]]`

which is completely weird, because user can't specify several passwords. You can fix it in any other desired way.

Also:
> `--lock`
> If this is present, the root account is locked by default. That is, the root user will not be able to login from the console.
> New in version Fedora8.

here is no info about **the password argument is not required for when --lock is specified**. It is not clear from this text.

atodorov:
@Frodox - thanks for the pointer, I see your pain now. #129 adds a partial fix for the later issue. As for the password argument being rendered as  `[<password> [<password> ...]]` this comes from argparse I think. I'll see if we can do something about it but not sure.

Frodox:
> As for the password argument being rendered as `[<password> [<password> ...]]` this comes from argparse I think. I'll see if we can do something about it but not sure.

I think, `[<password>]` will be just fine (+docs, in which cases it is not required)

clumens:
I'd like to get this merged - can someone tell me the current status of this PR?  Is there anything left to be done on it before it can be merged?

Frodox:
@clumens looks like @atodorov wants to fix it in #129 way :)

clumens:
Closed in favor of #129.  Thanks for the patch anyway.

PR #121, opened 2016-10-26
Remove auto generated files from docs folder

jkonecny12:
These files are generated from `make docs` and they shouldn't be tracked by git.


clumens:
NAK, please see 5f19e2fc5936e8a740547306bc6f6bf87337439e

jkonecny12:
Wouldn't it be better to have these files in a separate repository? For me it is confusing to have these files here. I personally wrote the documentation here first and then run make docs which erased what I wrote. When you don't know the structure it's not easy task to distingue generated files from "for edit" files.
If you must synchronize and commit the generated files from time to time, then it's not so harder to do this in another repository with just documentation, this could be a git subtree for example.

 I know everything has it's burden. If nothing is usable for you here, please rename the file to be crystal clear that these files are automatically generated, that would help a lot.

M4rtinK:
> Wouldn't it be better to have these files in a separate repository?

I second that!

> For me it is confusing to have these files here. I personally wrote the documentation here first and then run make docs which erased what I wrote. When you don't know the structure it's not easy task to distingue generated files from "for edit" files.
> If you must synchronize and commit the generated files from time to time, then it's not so harder to do this in another repository with just documentation,

Yeah - I also like that idea. And we could use a Jenkins job to the repository automatically once the "parent" pykickstart repository changes - we already use the corresponding hooks to trigger test runs and this would not be much different. So no manual interaction would be needed once setup.

> this could be a git subtree for example.

I'm quite sure the content of the subtree would still show up in `git grep` and similar - which might be confusing. You want to search for the `network` command and it returns a bunch of autogenerated stuff (we have similar issues with auto-generated docs in the Anaconda repository and it is rather annoying).

So I would recommend separate repository (`pykickstart-docs` or `pykickstart-readthedocs` ?)

> 
> I know everything has it's burden. If nothing is usable for you here, please rename the file to be crystal clear that these files are automatically generated, that would help a lot.

Yeah, that would certainly help for the time being (if possible).

atodorov:
> > Wouldn't it be better to have these files in a separate repository?
> I second that!

I will also add a section into README and/or docs explaining how the documentation is structured and built so people don't get confused.

M4rtinK:
> Wouldn't it be better to have these files in a separate repository?
>         I second that!
> 
> I will also add a section into README and/or docs explaining how the documentation is structured and built so people don't get confused.

Sounds good & thanks in advance! :)

clumens:
I really don't want a second repo for docs.  I don't want to make this more complicated than it already is.  I don't want a second place to have to look for things.  I'd prefer us to figure out a way to work with readthedocs, which I am honestly starting to view as having been a bad idea.

M4rtinK:
> I really don't want a second repo for docs. I don't want to make this more complicated than it already is. I don't want a second place to have to look for things. I'd prefer us to figure out a way to work with readthedocs,

I think running the makefile/the code makefile runs to auto-generate the rst files from the config.py file should work - I [found some information about it on Stack Overflow](http://stackoverflow.com/a/39619152), namely:


`The configuration file is executed as Python code at build time (using execfile(), and with the current directory set to its containing directory), and therefore can execute arbitrarily complex code. Sphinx then reads simple names from the file’s namespace as its configuration.`

This combined with the fact that the docs generation runs in a virtual environment where we [can request arbitrary packages from PyPy](http://read-the-docs.readthedocs.io/en/latest/faq.html#my-project-isn-t-building-with-autodoc) should make it possible.

Also it looks like [RTD could also be told to `setup.py install` first](http://read-the-docs.readthedocs.io/en/latest/builds.html#understanding-what-s-going-on):

`If you have the Install Project option enabled, we will run setup.py install on your package, installing it into a virtual environment.`

That might be another point where to hook the rst file generation - possibly cleaner than doing it in config.py & actually a quite sensible thing to do for a `setup.py install` call (eq. install with auto generated docs).

> which I am honestly starting to view as having been a bad idea.

I agree in an unnecessary big pain in the ass if a given project does not fit their (quite narrow) expected configuration. But I still really like the general idea and think it is worth it - docs auto generated & kept up to date from source repos, that are hosted in a single easily to find place.
 I also think it can help with discoverability in general as more and more projects host documentation on Read the Docs (Django, Gluster, etc.) - kinda like with GitHub for project and source code.

clumens:
Martin, I remember us talking a while back about approaches that could be taken to solve this problem.  Did we decide on anything?  I'd like to get something merged to take care of this problem and get this PR closed.

M4rtinK:
I'm currently working on getting the Anaconda automatically generated documentation online on readthedocs and I plan to do it in a way that can be also used for other related projects such as Pykickstart, Lorax, etc.

So far it doesn't seem like there is any simple & robust way of making readthedocs run all the docs generation steps we need, so I've went the separate-docs repo way.

I've created the [anaconda-docs](https://github.com/rhinstaller/anaconda-docs) which contains a simple Python script which updates a local Anaconda checkout & generates new docs from the once run. The script can also commit (only the generated docs, not the Anaconda checkout) the repo & push changes to origin via command line options. I've also removed the autogenerated docs files from the main anaconda repo.

The idea is that a job in our Jenkins instance will trigger this script when the Anaconda repo changes (or even just once a day), the script will update the generated docs, commit them and pus them. According to Jiri it should be easy to do. Readthedocs will then pick up the changes in the repo and update it's Anaconda docs instance.

If you would be fine with this solution also for Pykickstart I can setup the same (a repo for pykickstart auto-generated docs & a Jenkins job to update the docs) also for Pykickstart.

M4rtinK:
OK, scratch that - turn's out it **can** be done directly without an intermediate project:
https://github.com/rhinstaller/pykickstart/pull/132

clumens:
#132 does everything this one does and more, so closing.  But thanks for bringing this problem up.

M4rtinK:
Also thanks to you & all the others who did all the work on the actual sphinx docs generation - otherwise I would have nothing to work with. :)

PR #120, opened 2016-10-26
Master add noninteractive

jkonecny12:
Reworked PR https://github.com/rhinstaller/pykickstart/pull/111 from `pykickstart-2` for master.

PR #122, opened 2016-10-26
Remove duplicated kwargs.get displayMode variable

jkonecny12:
This is already in `FC3_DisplayMode` class so there is no need to have this in `F26_DisplayMode` class.

Introduced in PR https://github.com/rhinstaller/pykickstart/pull/111 . As far as I understand this, it shouldn't be used twice in the same command.


clumens:
Does #120 also need this fix?

jkonecny12:
Nope, I found this bug when I did that PR. So it's already fixed there.

PR #123, opened 2016-10-26
fix markup a bit (add ``)

Frodox:
Fix markup a bit in arguments and path used in descriptions


clumens:
Pushed, thanks for the patch.

PR #129, opened 2016-11-11
rootpw: document that password isn't required with --lock

atodorov:
Closes #119 

Also for the `sshpw` part note that the documentation says:
```
---sshkey 
If this is used then the <password> string is interpreted as an ssh key value.
```

which also means we can have a space character. Indeed there is also a test for that:
```
        self.assert_parse("sshpw --username=someguy --sshkey a ssh key with spaces", "sshpw --username=someguy --sshkey a ssh key with spaces\n")
```

It is unfortunate that the `<password>` argument is rendered as `[password [password ...]]` in the docs but that is a minor inconvenience given that the behavior is fully documented otherwise.

@Frodox, @clumens @jkonecny12. I propose to merge this one in favor of #119. Please review.

### DIFF
--- a/pykickstart/commands/rootpw.py
+++ b/pykickstart/commands/rootpw.py
@@ -55,7 +55,7 @@ class FC3_RootPw(KickstartCommand):
         op = KSOptionParser(prog="rootpw", description="""
                             This required command sets the system's root
                             password.""", version=FC3)
-        op.add_argument('password', metavar='<password>', nargs='*', version=FC3,
+        op.add_argument('password', metavar='<password>', nargs='?', version=FC3,
                         help="The desired root password.")
         op.add_argument("--iscrypted", dest="isCrypted", action="store_true",
                         default=False, version=FC3, help="""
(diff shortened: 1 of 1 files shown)


pykickstart/commands/rootpw.py
```diff
@@ -107,7 +106,8 @@ def _getParser(self):
                         version=F8, help="""
                         If this is present, the root account is locked by
                         default. That is, the root user will not be able to
-                        login from the console.""")
+                        login from the console. When this option is present
+                        the <password> argument is not required.""")
```
Frodox: may be `<password>` ?

Frodox:
Do `nargs='?'` will render like `[<password>]` ?

Looks good for me.

clumens:
This appears to be in conflict with #119.  Which should I take?

atodorov:
@clumens close #119 and merge this one.

atodorov:
> Do nargs='?' will render like [<password>] ?
@Frodox yes

pykickstart/commands/rootpw.py
atodorov: @Frodox I think what you're trying to tell me got eaten by markdown. Do you mean to use backticks to escape the "<password>" string so that it is rendered differently than the rest when RTD builds the HTML ?

PR #130, opened 2016-11-28
Remove mypy checks.

dashea:
Mypy is a moving target that really hasn't done anything interesting for
pykickstart, so remove the checks instead of trying to keep up with it.

The type annotations can stay since they aren't hurting anything. Maybe
someone will find them useful.


jkonecny12:
If we leave the type annotation they will be outdated and unusable later on. Also we have them in the git if we really want them in the future.

dashea:
Removed the type information. Also removed 'pip install mypy-lang' from .travis.yml since I missed that the first time.

PR #132, opened 2016-11-29
Run the docs makefile during RTD build

M4rtinK:
Run the docs makefile to generate intermediate documentation files
needed for a successful Read the Docs build.

This is achieved by running the make file from the conf.py script
which is executed during the RTD build. The second step needed is
is to visit advanced settings for the corresponding RTD project and:
- enable the "Install your project inside a virtualenv using setup.py install" option
- set the "Requirements file:" field to "docs/rtd-requirements.txt"

Also remove the pre-generated rst files, because these automatically generated
during the RTD build and no longer need to be stored in the repo.


M4rtinK:
For reference, this is my testing project building from my pykickstart fork:
https://readthedocs.org/projects/pykickstart-test/

As you can see from the build log, it does run the makefile:
https://readthedocs.org/projects/pykickstart-test/builds/4730412/

And the resulting docs look fine as far as I can tell, for example:
http://pykickstart-test.readthedocs.io/en/latest/kickstart-docs.html#chapter-2-kickstart-commands-in-fedora

PR #134, opened 2016-12-13
docs: Note under %include that most sections don't do merging

cgwalters:
This caught us when trying to have our Vagrant box inherit from the
OpenStack/cloud one and both having a `services` section.


clumens:
Merged, thanks.

PR #139, opened 2017-02-09
Remove unnecessary help docstrings

a-roussos:
In the online documentation for the `bootloader` command
(Fedora: http://pykickstart.readthedocs.io/en/latest/kickstart-docs.html#bootloader
RHEL: http://pykickstart.readthedocs.io/en/latest/kickstart-docs.html#id4)
under the `--md5pass` optional argument, the text "If using GRUB [...] already be encrypted." appears twice:
md5pass


clumens:
Hm, I think this is just a symptom of a larger problem.  The --md5pass command exists and then gets re-added later to change something about it.  The help text from the second add should be overriding the first, not being added to it.

PR #159, opened 2017-05-20
More documentation for bypassing the bootloader

amtlib-dot-dll:
Documentation about the difference between `--disabled` and `--location=none` according to https://bugzilla.redhat.com/show_bug.cgi?id=1222132#c2 and https://bugzilla.redhat.com/show_bug.cgi?id=1074522

### DIFF
--- a/pykickstart/commands/bootloader.py
+++ b/pykickstart/commands/bootloader.py
@@ -112,7 +112,15 @@ class FC3_Bootloader(KickstartCommand):
                         are the following: mbr (the default), partition
                         (installs the boot loader on the first sector of the
                         partition containing the kernel), or none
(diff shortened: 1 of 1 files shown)


clumens:
I'd like documentation to be useful independent of network access - I don't really like linking to a bug report in the documentation.  Also note that the bootloader command is useful for multiple platforms.  If you can take care of those two things, I'll merge the result.

amtlib-dot-dll:
Okay, I have done some modification accordingly. Thank you for your advice.

clumens:
Thanks for the updated - I squashed these into a single patch and merged to master.  I'll grab them for the pykickstart-2 branch as well.

PR #171, opened 2017-08-21
network: add network --bindto option

rvykydal:
By default ifcfgs/connections are bound to device name in anaconda (DEVICE).
This option allows to bind to mac address (HWADDR).

Related: rhbz#1328576

### DIFF
--- a/pykickstart/commands/network.py
+++ b/pykickstart/commands/network.py
@@ -15,10 +15,10 @@
 # Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.  Any Red Hat
 # trademarks that are incorporated in the source code or documentation are not
 # subject to the GNU General Public License and may only be used or replicated
-# with the express permission of Red Hat, Inc. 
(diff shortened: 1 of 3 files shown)


pykickstart/commands/network.py
```diff
@@ -667,4 +676,9 @@ def parse(self, args):
                     msg = formatErrorMsg(self.lineno, msg=_("Bad format of --bridgeopts, expecting key=value options separated by ','"))
                     raise KickstartValueError(msg)
 
+        if retval.bindto == BIND_TO_MAC:
+            if retval.vlanid and not retval.bondopts:
+                msg = formatErrorMsg(self.lineno, msg=_("--bindto=%s is not supported for this type of device" % BIND_TO_MAC))
```
clumens: The string substitution here should take place outside the call to _().  That is:
`msg=_("--bindto=%s...") % BIND_TO_MAC`

rvykydal: Oops, I did it again.

PR #178, opened 2017-10-04
Add a new 'mount' command

vpodzime:


pykickstart/commands/mount.py
```diff
@@ -0,0 +1,168 @@
...
+        if not other:
+            return False
+
+        return self.mount_point == other.mount_point
+
+    def __ne__(self, other):
+        return not self == other
+
+    def _getArgsAsStr(self):
+        retval = ""
```
jkonecny12: I guess you can remove this line.

vpodzime: Good catch!

clumens:
How does this fit in with the existing partitioning commands, especially where --noformat is involved?

vpodzime:
> How does this fit in with the existing partitioning commands, especially where --noformat is involved?

This is a simpler way,  it doesn't require the whole storage stack to be described in the kickstart (*raid*, *volgroup*, *logvol*,...) and it enables support for devices the installer (and/or *blivet*) doesn't fully understand. Any block device is fine for this command no matter where it came from. Last but not least, this is a base for the new way to configure storage in anaconda text mode -- users do whatever they want with their storage and then they just assign mount points to block devices. See https://github.com/vpodzime/anaconda/commits/mount_assign for details.

clumens:
Alright, could you make that more clear in the documentation then?  I feel like there's a fair bit of potential for confusion on the user side of things here - like, when would I use one method over the other?  Also, we'll need a patch for the master branch, too (which I really need to get around to just making the branch builds come from).  Thanks!

vpodzime:
> Alright, could you make that more clear in the documentation then? I feel like there's a fair bit of potential for confusion on the user side of things here - like, when would I use one method over the other?

Good point, done!

vpodzime:
> Also, we'll need a patch for the master branch, too (which I really need to get around to just making the branch builds come from).

Will do later today or tomorrow. But could we please move on with this ASAP? The plan is to make this available in Fedora 27 as a new experimental feature.

pykickstart/commands/mount.py
```diff
@@ -0,0 +1,165 @@
...
+        elif self.handler.logvol.seen:
+            conflicting_command = "logvol"
+        elif hasattr(self.handler, "reqpart") and self.handler.reqpart.seen:
+            conflicting_command = "reqpart"
+
+        if conflicting_command:
+            # allow for translation of the error message
+            errorMsg = _("The '%s' and 'mount' commands can't be used at the same time") % \
+                         conflicting_command
+            raise KickstartParseError(formatErrorMsg(self.lineno, msg=errorMsg))
```
poncovka: This handles only the case when the conflicting command is used before `mount`, no? What if it is used after `mount`?

clumens:
Yeah, there's no need to wait for a master patch before merging this.  I'll deal with each branch separately.

pykickstart/commands/mount.py
vpodzime: That's a good point. @clumens, do I need to add new version of all the ``autopart``, ``part``, ``raid``, ``volgroup``, ``logvol`` and ``reqpart`` comands just to make sure their ``parse()`` methods check that ``mount`` wasn't used together with them?

clumens: Nah, I don't think that is necessary.  Adding the test to the older versions of commands doesn't change their behavior.

vpodzime: Done.

M4rtinK:
I'm ready to handle the Anaconda part (merging the Anaconda PR & Anaconda F27 build) and a combined Anaconda + Pykickstart Bodhi update once this PR is merged and there is a new Pykickstart build for F27 with it.

PR #190, opened 2017-10-23
Add a new 'mount' command

vpodzime:

(cherry picked from commit
5d0f4503c56b54f42b9bed7a076843dfb2c3da76)

Related: rhbz#1450922


vpodzime:
This is a RHEL 7 version of #178

PR #191, opened 2017-10-23
Fix which Autopart class checks for the 'mount' command

vpodzime:
It should be the F20_Autopart version from which new versions of
the class used in Fedora 20+ and RHEL 7 inherit, not the archaic
RHEL6 version of the class which can never meet the 'mount'
command.


vpodzime:
Unfortunately, I have only spotted this mistake today while porting the *mount* command support to *rhel7-branch*.

PR #192, opened 2017-10-23
Add a new 'mount' command

vpodzime:

Ported and squashed commits:
  5d0f4503c56b54f42b9bed7a076843dfb2c3da76
  0a328f7f787a83ed06796368fa72c0da90b81047
from the pykickstart-2 branch.


vpodzime:
A *master* branch version of #178

PR #196, opened 2017-11-06
Commands for specifying base repo are mentioned in docs

jkonecny12:
Commands which are required for specifying base repo weren't mentioned in docs. So users can get easily confused that without these commands the closest mirror will be used.

### DIFF
--- a/docs/kickstart-docs.rst
+++ b/docs/kickstart-docs.rst
@@ -83,6 +83,10 @@ kickstart file:
    user would be prompted during a typical installation. Once the answer
    is given, the installation will continue unattended unless it finds
    another missing item.
+-  One installation source command from the list of commands in the ``method``
+   proxy command must be specified for the fully automated kickstart
+   installation. This is required even for Fedora -- the closest mirror can't
+   be chosen by the kickstart file.
 -  Lines starting with a pound sign (#) are treated as comments and are
    ignored.
 -  If deprecated commands, options, or syntax are used during a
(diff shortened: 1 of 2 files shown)


docs/kickstart-docs.rst
```diff
@@ -83,6 +83,9 @@ kickstart file:
    user would be prompted during a typical installation. Once the answer
    is given, the installation will continue unattended unless it finds
    another missing item.
+-  Commands ``url``, ``nfs``, ``cdrom`` or ``harddrive`` are required for
```
poncovka: Also `liveimg` and `hmc`. In general, every command specified in the `method` command. 

Maybe the list of commands could be generated in the `method`'s help and here it could be just pointed out that we require one of the installation methods specified at `method`

jkonecny12: Hmm interesting idea. I'll look on that.

jkonecny12:
**UPDATED:**
* List supported commands in the ``method`` proxy command.
* Change text to refer list in a ``method`` proxy command.

docs/kickstart-docs.rst
```diff
@@ -83,6 +83,9 @@ kickstart file:
    user would be prompted during a typical installation. Once the answer
    is given, the installation will continue unattended unless it finds
    another missing item.
+-  Commands mentioned in the ``method`` proxy command are required for
```
poncovka: It looks like all of these commands are required.

jkonecny12:
**UPDATED:**
* Explicitly tell that one command must be used from the list.

docs/kickstart-docs.rst
jkonecny12: Huh, true..

PR #203, opened 2017-12-15
master: firewall: add --use-system-defaults arg to command

dustymabe:
Needed for [1] where we would like to include firewalld
and configure firewalld in Atomic Host (in the ostree) and
have Anaconda leave the delivered "defaults" in place.

[1] https://pagure.io/atomic-wg/issue/401

### DIFF
--- a/pykickstart/commands/firewall.py
+++ b/pykickstart/commands/firewall.py
@@ -17,7 +17,7 @@
 # subject to the GNU General Public License and may only be used or replicated
 # with the express permission of Red Hat, Inc.
 #
-from pykickstart.version import FC3, F9, F10, F14, F20
+from pykickstart.version import FC3, F9, F10, F14, F20, F28
 from pykickstart.base import KickstartCommand
 from pykickstart.options import ExtendAction, ExtendConstAction, KSOptionParser, commaSplit
 
@@ -240,3 +240,25 @@ class F20_Firewall(F14_Firewall):
             return retval + "%s\n" % svcstr
         else:
             return retval
+
+class F28_Firewall(F20_Firewall):
+    def __init__(self, writePriority=0, *args, **kwargs):
+        F20_Firewall.__init__(self, writePriority, *args, **kwargs)
+        self.use_system_defaults = kwargs.get("use_system_defaults", None)
+
+    def _getParser(self):
+        op = F20_Firewall._getParser(self)
+        op.add_argument("--use-system-defaults", dest="use_system_defaults",
+                        action="store_true", default=False, version=F28, help="""
+                        Don't configure the firewall at all. This instructs anaconda
+                        to do nothing and allows the system to rely on the defaults
+                        that were provided with the package or ostree.  If this option
+                        is used with other options then all other options will be
+                        ignored.""")
+        return op
+
+    def __str__(self):
+        if self.use_system_defaults:
+            return "# Firewall configuration\nfirewall --use-system-defaults\n"
+        else:
+            return F20_Firewall.__str__(self)
--- a/tests/commands/firewall.py
+++ b/tests/commands/firewall.py
@@ -20,7 +20,7 @@
 
 import unittest
 from tests.baseclass import CommandTest
-from pykickstart.commands.firewall import FC3_Firewall, F10_Firewall, F20_Firewall
+from pykickstart.commands.firewall import FC3_Firewall, F10_Firewall, F20_Firewall, F28_Firewall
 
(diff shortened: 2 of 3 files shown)


pykickstart/commands/firewall.py
```diff
@@ -218,14 +218,23 @@ class F20_Firewall(F14_Firewall):
...
         F14_Firewall.__init__(self, writePriority, *args, **kwargs)
         self.remove_services = kwargs.get("remove_services", [])
+        self.use_system_defaults = kwargs.get("use_system_defaults", None)
 
     def _getParser(self):
         op = F14_Firewall._getParser(self)
         op.add_argument("--remove-service", dest="remove_services", help="",
                         action=ExtendAction, type=commaSplit, version=F20)
+        op.add_argument("--use-system-defaults", dest="use_system_defaults",
+                        version=F20, help="""
```
cgwalters: Shouldn't this have a higher `version=`?

dustymabe: I was mostly matching the `--remove-service` option that was already there. Depending on this review I might have to make a whole new subclass for F28, but I don't know what the convention is.

clumens: This is probably a little confusing and could be better documented... somewhere.  In this context, the version= parameter is used to generate documentation.  For your new option, you would use version=F28 (or whatever).

The general idea is that the version is whatever release of Fedora or RHEL it first appeared in.  You would want to make a new subclass here, yes.  I'd be happy to help with that if you need pointers.

dustymabe: I can take a stab at it. would I have to create a new subclass for the other branches as well?

pykickstart/commands/firewall.py
```diff
@@ -240,3 +240,23 @@ def __str__(self):
...
+                        Don't configure the firewall at all. This instructs anaconda
+                        to do nothing and allows the system to rely on the defaults
+                        that were provided with the package or ostree.""")
+        return op
+
+    def __str__(self):
+        if self.use_system_defaults:
+            return "# Firewall configuration\nfirewall --use-system-defaults\n"
+        else:
+            return F20_Firewall.__str__(self)
```
clumens: This is going to trip up whoever has to make a new class that inherits from this one and needs to do their own __str__ method, but I don't see that we can do much about that right now.

tests/commands/firewall.py
```diff
@@ -124,5 +125,12 @@ def runTest(self):
         self.assert_parse("firewall --service=mdns --remove-service=dhcpv6-client --remove-service=ssh",
                           "firewall --enabled --service=mdns --remove-service=dhcpv6-client,ssh\n")
 
+class F28_TestCase(F20_TestCase):
+    def runTest(self):
+        F20_TestCase.runTest(self)
+
+        # use-system-defaults
+        self.assert_parse("firewall --use-system-defaults\n")
+
```
clumens: I'd like to see an additional test for when you use `--use-system-defaults` and some other option, to verify that part works.  That could also help out with the other comment I left.

dustymabe: passing something like `firewall --use-system-defaults  --enabled --service=ssh,smtp,ftp ` probably doesn't make sense, though. the `--use-system-defaults` is literally like a short circuit that says 'do nothing'. Do you agree? What are you looking for here?

jkonecny12: If it doesn't make sense than it should be an exception.

clumens: That's one possibility - yeah.  An exception could be raised.

I know it doesn't make sense.  What I'm looking for is a test to make sure that the thing that doesn't make sense is being handled correctly.

dustymabe: ok, so you want me to raise an exception when someone provides `--use-system-defaults` and another option?

dustymabe: I looked into this quite a bit today. I really tried to figure out how I could use argparse [mutual exclusion](https://docs.python.org/2/library/argparse.html#mutual-exclusion) to achieve this goal but unfortunately it seems the way things are wired up in anaconda (i.e. the overiding of the `add_argument` function) does not lend itself well to this. Do you know of any other examples where exclusivity is required in anaconda?

Within this very same module one can pass `--disable --enable` and anaconda will parse it fine.

dustymabe: bump

dustymabe:
ok rebased and updated based on code review comments. I think I did everything.

PR #204, opened 2017-12-15
pykickstart-2: firewall: add --use-system-defaults arg to command

dustymabe:
Needed for [1] where we would like to include firewalld
and configure firewalld in Atomic Host (in the ostree) and
have Anaconda leave the delivered "defaults" in place.

[1] https://pagure.io/atomic-wg/issue/401

### DIFF
--- a/pykickstart/commands/firewall.py
+++ b/pykickstart/commands/firewall.py
@@ -233,3 +233,20 @@ class F20_Firewall(F14_Firewall):
             return retval + "%s\n" % svcstr
         else:
             return retval
+
+class F28_Firewall(F20_Firewall):
+    def __init__(self, writePriority=0, *args, **kwargs):
+        F20_Firewall.__init__(self, writePriority, *args, **kwargs)
+        self.use_system_defaults = kwargs.get("use_system_defaults", None)
+
+    def _getParser(self):
+        op = F20_Firewall._getParser(self)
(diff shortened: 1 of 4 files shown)


clumens:
I'd like to see the test case from the master version ported here.

dustymabe:
the reason I didn't add a test case is because I did a `grep -R F20_Firewall ./` and didn't find any test files so I assumed there wasn't one. It just turns out the **F20_TestCase** among other things was never done on this branch. Do you mind updating it so that the only thing I need to add is the **F28_TestCase** ?

dustymabe:
> Do you mind updating it so that the only thing I need to add is the F28_TestCase ?

bump

clumens:
That would be fine.

dustymabe:
> That would be fine.

any chance you could do that today and then I can update this PR so we can get it in soon?

clumens:
Done.

dustymabe:
ok updated this branch with tests. let me run a test real quick to make sure I got everything.. also for some reason the PR on the master branch you merged ended up with a different commit hash than what I had locally even though the contents are the same. If the same happens for this branch then my **cherry picked from xxxx** message will be wrong in the rhel7 branch

dustymabe:
ok looks mostly good from testing with atomic host

PR #206, opened 2018-01-12
Pykickstart 2 remove btrfs support

rvykydal:


rvykydal:
Just added rhbz tag to commit message

PR #207, opened 2018-02-08
Copy txt files from _build folder on make local call

jkonecny12:
Without this you won't be able to do a build with make local because `kickstart-docs.txt` file is used to create an rpm but expected in docs folder and not in `docs/_build` folder.


jkonecny12:
We need this to make daily builds.
We can also switch to `make archive` but I don't think we want to loose time for tests when they will also run in a build process.

PR #210, opened 2018-02-14
Master remove btrfs command

rvykydal:


rvykydal:
This should obsolete
https://github.com/clumens/pykickstart/pull/206

pykickstart/handlers/rhel8.py
```diff
@@ -0,0 +1,119 @@
...
+        "user": commands.user.F24_User,
+        "vnc": commands.vnc.F9_Vnc,
+        "volgroup": commands.volgroup.RHEL8_VolGroup,
+        "xconfig": commands.xconfig.F14_XConfig,
+        "zerombr": commands.zerombr.F9_ZeroMbr,
+        "zfcp": commands.zfcp.F14_ZFCP,
+    }
+
+    dataMap = {
+        "BTRFSData": commands.btrfs.F23_BTRFSData,
```
poncovka: Should this line stay here?

clumens: I don't think we've ever had a case where we removed a command that had associated data classes, so we're in uncharted territory here.  I don't really know what the right thing to do is, but I lean towards leaving it.  We'll still recognize the "btrfs" command for purposes of not breaking everyone's stuff, so we should probably leave the data class around for the same reason.  We should probably also just have some way of marking it as deprecated.

clumens: Obviously in this patch, it should stay - this is just adding the structure to be modified later.

poncovka: Ok.

PR #213, opened 2018-02-21
Add the module command

M4rtinK:
Add the module command used for enabling modules.

The module command takes two options:
--name - set the name of the module to enable
--stream - set the stream name to of the module to enable


pykickstart/commands/enablemodule.py
```diff
@@ -0,0 +1,100 @@
...
+    def __str__(self):
+        retval = ""
+
+        for enablemodule in self.enablemoduleList:
+            retval += enablemodule.__str__()
+
+        return retval
+
+    def _getParser(self):
+        op = KSOptionParser(prog="enablemodule", version=F28, description="""
```
poncovka: I think that the command should be named just `module`. In the future, the command might not always enable the module (but disable it for example).

pykickstart/commands/enablemodule.py
```diff
@@ -0,0 +1,100 @@
...
+                            Enable a module.""")
```
poncovka: Please, extend the documentation a little. It is not clear what is a module.

pykickstart/commands/enablemodule.py
M4rtinK: Good point, let's do that!

pykickstart/commands/enablemodule.py
M4rtinK: Indeed, there is already a lot of things one can call "module".

pykickstart/commands/module.py
```diff
@@ -0,0 +1,115 @@
...
+
+    A module is defined by a unique name and a stream id, where single module can
+    (and usually has) multiple available stream.
+
+    Streams will in most cases corresponds to stable releases of the given software
+    components (such as Node.js, Django, etc.) but there could be also other use cases,
+    such as a raw master stream or streams corresponding to an upcoming stable stream.
+
+    For more information see the Fedora modularity initiative documentation:
+    https://docs.pagure.org/modularity/
```
poncovka: This looks good, just move to the description of the command, so it will be generated in the documentation, please.

pykickstart/commands/module.py
```diff
@@ -0,0 +1,115 @@
...
+                        Name of the module stream to enable.""")
+
+        return op
+
+    def parse(self, args):
+        (ns, extra) = self.op.parse_known_args(args=args, lineno=self.lineno)
+
+        if len(extra) > 0:
+            msg = _("The enable module command does not take position arguments!")
+            raise KickstartParseError(formatErrorMsg(self.lineno, msg=msg), lineno=self.lineno)
```
poncovka: Function `formatErrorMsg` is deprecated now, use `KickstartParseError(msg, lineno=self.lineno)`.

clumens:
Where are we on this?  I'd be happy to click the merge button if the disclaimer from the first comment has been addressed.  I'd hate for this to get lost.

M4rtinK:
> Where are we on this? I'd be happy to click the merge button if the disclaimer from the first comment has been addressed. I'd hate for this to get lost.

Work in Anaconda & DNF is ongoing and progressing well (it's already possible to enable & install modules with the WiP code) and while it's unlikely the module command syntax will change dramatically, I'd like to wait with this PR for the others (Anaconda & DNF) to be declared ready as well.

I also still need to apply the suggestions from Vendy.

In any case, rest assured I'll let you know at once once this PR can be merged. :)

pykickstart/commands/module.py
M4rtinK: Fixed, thanks! :)

pykickstart/commands/module.py
M4rtinK: Fixed as well.

M4rtinK:
The module command seem to be working fine so far and further changes are unlikely, so let's merge this now to reduce the amount of modularity related stuff still in flight.

PR #215, opened 2018-04-06
Rhel8 branch add nvdimm command

rvykydal:


rvykydal:
This is the same PR as https://github.com/clumens/pykickstart/pull/214 which is for master.

rvykydal:
@clumens - could you please take a look?
It has been already reviewed by anaconda folks in https://github.com/clumens/pykickstart/pull/214 (the same PR for master)

clumens:
I just pushed a patch that updates a test case for writePriority for the nvdimm command on master.  That'll need to be done here too, but I'll take care of it.  Merging, thanks.

PR #230, opened 2018-06-21
Add the syspurpose command

M4rtinK:
The syspurpose ("system purpose") command makes it possible for the
user to describe how the given system is intended to be used after
the installation.

This information can then be used to properly subscribe the system
to the correct entitlements (in cases subscribing a system is possible).

### DIFF
--- a/pykickstart/commands/syspurpose.py
+++ b/pykickstart/commands/syspurpose.py
@@ -1,0 +1,89 @@
+#
+# Martin Kolman <mkolman@redhat.com>
+#
+# Copyright 2018 Red Hat, Inc.
+#
+# This copyrighted material is made available to anyone wishing to use, modify,
+# copy, or redistribute it subject to the terms and conditions of the GNU
+# General Public License v.2.  This program is distributed in the hope that it
+# will be useful, but WITHOUT ANY WARRANTY expressed or implied, including the
+# implied warranties of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+# See the GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along with
+# this program; if not, write to the Free Software Foundation, Inc., 51
+# Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.  Any Red Hat
+# trademarks that are incorporated in the source code or documentation are not
+# subject to the GNU General Public License and may only be used or replicated
+# with the express permission of Red Hat, Inc.
+#
+
+from pykickstart.base import KickstartCommand
+from pykickstart.errors import KickstartParseError
+from pykickstart.options import KSOptionParser
+from pykickstart.version import RHEL8
+
+from pykickstart.i18n import _
+
+class RHEL8_Syspurpose(KickstartCommand):
+    removedKeywords = KickstartCommand.removedKeywords
+    removedAttrs = KickstartCommand.removedAttrs
+
+    def __init__(self, writePriority=0, *args, **kwargs):
+        KickstartCommand.__init__(self, writePriority, *args, **kwargs)
+        self.op = self._getParser()
+
+        self.role = kwargs.get("role", None)
+        self.sla = kwargs.get("sla", None)
+        self.usage = kwargs.get("usage", None)
+        self.addons = kwargs.get("addons", None)
+
+    def __str__(self):
+        retval = KickstartCommand.__str__(self)
+
+        if not retval and not any([self.role, self.sla, self.usage, self.addons]):
+            return ""
+
+        retval += '# Intended system purpose\nsyspurpose'
+        if self.role:
+            retval+=' --role="%s"' % self.role
+        if self.sla:
+            retval+=' --sla="%s"' % self.sla
+        if self.usage:
+            retval+=' --usage="%s"' % self.usage
+        if self.addons:
+            for addon in self.addons:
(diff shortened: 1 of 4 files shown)


pykickstart/commands/syspurpose.py
```diff
@@ -0,0 +1,82 @@
...
+            retval+=" --role=%s" % self.role
+        if self.sla:
+            retval+=" --sla=%s" % self.sla
+        if self.usage:
+            retval+=" --usage=%s" % self.usage
+
+        return retval
+
+    def _getParser(self):
+        op = KSOptionParser(prog="syspurpose", description="""
```
poncovka: The name `syspurpose` looks really weird. It might be better to use just `purpose` or we could start to support long command names, then it could be `system-purpose`.

pykickstart/commands/syspurpose.py
```diff
@@ -0,0 +1,82 @@
...
+        op.add_argument("--sla", metavar="<sla_name>", version=F29, required=False,
+                        help="""
+                        Name of the sla intended for the system.""")
+        op.add_argument("--usage", metavar="<usage>", version=F29, required=False,
+                        help="""
+                        The intended usage of the system.""")
+        return op
+
+    def parse(self, args):
+        (ns, extra) = self.op.parse_known_args(args=args, lineno=self.lineno)
```
poncovka: I think that there is a method that will parse all arguments, so you don't have to do the check later.

pykickstart/commands/syspurpose.py
```diff
@@ -0,0 +1,82 @@
...
+#
+# You should have received a copy of the GNU General Public License along with
+# this program; if not, write to the Free Software Foundation, Inc., 51
+# Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.  Any Red Hat
+# trademarks that are incorporated in the source code or documentation are not
+# subject to the GNU General Public License and may only be used or replicated
+# with the express permission of Red Hat, Inc.
+#
+
+from pykickstart.base import BaseData, KickstartCommand
```
clumens: For once, travis is telling us something useful.  The import of BaseData is unnecessary.

pykickstart/commands/syspurpose.py
clumens: There's a parse_args method and a parse_known_args method.  Beyond that, I'm not sure what you are referring to here.

pykickstart/commands/syspurpose.py
```diff
@@ -0,0 +1,82 @@
...
+                            correct subscription entitlement to the system.""", version=F29)
+        op.add_argument("--role", metavar="<role_name>", version=F29, required=False,
+                        help="""
+                        The intended role of the system.""")
+        op.add_argument("--sla", metavar="<sla_name>", version=F29, required=False,
+                        help="""
+                        Name of the sla intended for the system.""")
+        op.add_argument("--usage", metavar="<usage>", version=F29, required=False,
+                        help="""
+                        The intended usage of the system.""")
```
clumens: Could you expand what you mean by role, SLA (service level agreement?), and usage in the help just a little bit?  I'm not sure what any of them means in this context.  Just a sentence or so would be sufficient.

clumens:
Assuming the unnecessary import of BaseData is fixed, this can be merged.

M4rtinK:
> Assuming the unnecessary import of BaseData is fixed, this can be merged.

Should be now fixed. :)

clumens:
Still approved, then.

M4rtinK:
Some more changes:
- added the addons option
- improved multi word value handling & multi word value unit tests

pykickstart/handlers/f29.py
```diff
@@ -85,6 +85,7 @@ class F29Handler(BaseHandler):
         "snapshot": commands.snapshot.F26_Snapshot,
         "sshpw": commands.sshpw.F24_SshPw,
         "sshkey": commands.sshkey.F22_SshKey,
+        "syspurpose" : commands.syspurpose.F29_Syspurpose,
```
poncovka: Does it make sense to have this command in Fedora? Shoudn't you use a different handler?

tests/commands/syspurpose.py
```diff
@@ -0,0 +1,60 @@
...
+        self.assert_parse('syspurpose --role=foo --sla=bar --usage=baz --addons=a,b,c')
+        self.assert_parse('syspurpose --role=foo')
+        self.assert_parse('syspurpose --sla=bar')
+        self.assert_parse('syspurpose --usage=baz')
+        self.assert_parse('syspurpose --addons=a,b,c')
+        # just syspurpose without options is likely valid even though useless
+        self.assert_parse('syspurpose')
+
+        # multi word names
+        self.assert_parse('syspurpose --role="foo a" --sla="bar b c" --usage="baz foo" --addons="a b,c d,e"')
```
poncovka: Please, remove the `assert_parse` calls with only one argument. They are redundant.

clumens:
When poncovka is happy with this patch, I'm happy and you can merge it.  Thanks, everyone!

M4rtinK:
Some hopefully final updates:
- switch to the RHEL handler for the command
- as addons names can contain , and other weird characters we need to have one addon per `--addon` arguments & support `--addon` being used multiple times

M4rtinK:
Also opened a PR with syspurpose command rebased on the RHEL branch: https://github.com/clumens/pykickstart/pull/237

PR #238, opened 2018-08-28
Add missing trailing newline for syspurpose __str__ method

M4rtinK:
Without the trailing newline the next command will be written
on the same line, leading to ugly or even invalid kickstart file.

Related: rhbz#1612061


pykickstart/commands/syspurpose.py
```diff
@@ -54,6 +54,8 @@ def __str__(self):
         if self.addons:
             for addon in self.addons:
                 retval+=' --addon="%s"' % addon
+        # finally, add a newline
```
poncovka: Please, remove the comment.

PR #244, opened 2018-09-24
Fix typo in ksshell man page

scfc:


clumens:
Thanks for the patch, merged.

PR #245, opened 2018-09-24
Mention all man pages in all "SEE ALSO" sections

scfc:


clumens:
Thanks for the patch, merged.

PR #246, opened 2018-10-09
Document module stream installation

M4rtinK:
Describe how to install module streams & profiles, including
which combinations are valid and which not.

The group/module priority is also defined for ambiguous
cases (module wins).

### DIFF
--- a/pykickstart/sections.py
+++ b/pykickstart/sections.py
@@ -494,6 +494,38 @@ class PackageSection(Section):
                             with no additional characters (the dhcp line in the
                             example above is an individual package).
 
+                            The ``@`` prefix is also used to request installation
+                            of module streams in the following format::
+
+                                @<module name>:<stream name>/<profile name>
+
+                            Profile name is optional and multiple profiles can be
+                            installed by using multiple lines, one per profile.
+                            Stream name is only optional only if the given module
+                            has a default stream.
+
+                            If there are a module and a group named the same,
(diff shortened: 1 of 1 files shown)


pykickstart/sections.py
```diff
@@ -494,6 +494,38 @@ def _getParser(self):
...
+                            Profile name is optional and multiple profiles can be
+                            installed by using multiple lines, one per profile.
+                            Stream name is only optional only if the given module
+                            has a default stream.
+
+                            If there are a module and a group named the same,
+                            and no stream name and profile are specified,
+                            module will be selected instead of a group.
+
+                            Requesting one module more than one with different
```
bcl: I'm not sure what this line means.

pykickstart/sections.py
```diff
@@ -494,6 +494,38 @@ def _getParser(self):
...
+                                @<module name>:<stream name>/<profile name>
+
+                            Profile name is optional and multiple profiles can be
+                            installed by using multiple lines, one per profile.
+                            Stream name is only optional only if the given module
+                            has a default stream.
+
+                            If there are a module and a group named the same,
+                            and no stream name and profile are specified,
+                            module will be selected instead of a group.
```
bcl: How do you select the group then? Wouldn't it make more sense to select the group if there is a name conflict, and require the module to have the stream specified?

PR #247, opened 2018-10-15
Do not throw exception on deprecated ignoredisk --interactive

rvykydal:
The option was deprecated by
commit 67bde6b1bfe0a20d17a95620d56df96dfc805efej
but we don't want to abort the installation for existing kickstarts using the
option so loose the guard in parse method and add a deprecation warning.

If there is the --interactive option only, just issue the warning, do not raise
parse error. (These kickstarts used to be accepted.)

If there is also other option present in the command, modify the message of
raised exception with information that the --interactive option is no longer
supported. (These kickstarts used to raise the exception but the message should
be changed.)

Resolves: rhbz#1637872

### DIFF
--- a/pykickstart/commands/ignoredisk.py
+++ b/pykickstart/commands/ignoredisk.py
@@ -17,9 +17,11 @@
 # subject to the GNU General Public License and may only be used or replicated
 # with the express permission of Red Hat, Inc.
 #
+import warnings
+
 from pykickstart.version import FC3, F8, RHEL6, F29
 from pykickstart.base import KickstartCommand
-from pykickstart.errors import KickstartParseError
+from pykickstart.errors import KickstartParseError, KickstartDeprecationWarning
 from pykickstart.i18n import _
 from pykickstart.options import KSOptionParser, commaSplit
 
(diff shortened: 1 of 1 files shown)


rvykydal:
Note: we still don't have mechanism to pass warnings from modules (Storage module in this case) to anaconda/Boss so the message will be visible only in Storage module log and won't be exposed when using inst.ksstrict option.

pykickstart/commands/ignoredisk.py
```diff
@@ -151,7 +153,29 @@ class F29_IgnoreDisk(F14_IgnoreDisk):
...
+        if self.onlyuse:
+            howmany += 1
+        if self.interactive:
+            howmany += 1
+            warnings.warn(_("Ignoring deprecated option on line %s:  The ignoredisk command "
+                            "no longer supports --interactive option.  In future releases, "
+                            "this will result in a fatal error from kickstart. "
+                            "Please modify your kickstart file to remove the option.")
+                          % self.lineno, KickstartDeprecationWarning)
+        if howmany != 1:
```
jkonecny12: What about removing the `howmany` and change this line to:

```suggestion
        if sum(map(int, (self.ignoredisk, self.onlyuse, self.interactive))) != 1:
```

pykickstart/commands/ignoredisk.py
```diff
@@ -151,7 +153,29 @@ class F29_IgnoreDisk(F14_IgnoreDisk):
...
+        retval = FC3_IgnoreDisk.parse(self, args)
+
+        howmany = 0
+        if self.ignoredisk:
+            howmany += 1
+        if self.onlyuse:
+            howmany += 1
+        if self.interactive:
+            howmany += 1
+            warnings.warn(_("Ignoring deprecated option on line %s:  The ignoredisk command "
```
poncovka: I think that this warning should be already produced by the argument parser.

rvykydal: I am not able to find the warning anywhere in the logs. Maybe this holds only for deprecated commands?
BTW, the warning will be only displayed in storage module log as we don't forward kickstart parsing warnings from modules to anaconda (just errors). Which is perhaps ok for now.

pykickstart/commands/ignoredisk.py
rvykydal: Yes, howmany seems a bit funny to me, but I prefer to keep it consistent with other code.

jkonecny12: OK, just an idea :). You know in deep of my hearth I like functional programming :D.

PR #250, opened 2018-12-21
Add options for ssl certs on url and repo commands

larskarlitski:
This adds --sslclientcert, --sslclientkey, and --sslcacert options to
the `url` and `repo` commands. They mirror the options in dnf.conf.


dcantrell:
This change needs updated test cases to account for the new options for 'url' and the new command revision.

pykickstart/commands/repo.py
```diff
@@ -168,6 +168,23 @@ def _getArgsAsStr(self):
...
+        F27_RepoData.__init__(self, *args, **kwargs)
+        self.sslcacert = kwargs.get("sslcacert", None)
+        self.sslclientcert = kwargs.get("sslclientcert", None)
+        self.sslclientkey = kwargs.get("sslclientkey", None)
+
+    def _getArgsAsStr(self):
+        return F27_RepoData._getArgsAsStr(self) + \
+            " --sslcacert=%s" % self.sslcacert + \
+            " --sslclientcert=%s" % self.sslclientcert + \
+            " --sslclientkey=%s" % self.sslclientkey
```
poncovka: This code doesn't handle `None`.

pykickstart/commands/url.py
```diff
@@ -241,3 +241,57 @@ def _getParser(self):
...
+    removedAttrs = F27_Url.removedAttrs
+
+    def __init__(self, *args, **kwargs):
+        F27_Url.__init__(self, *args, **kwargs)
+        self.sslcacert = kwargs.get("sslcacert", None)
+        self.sslclientcert = kwargs.get("sslclientcert", None)
+        self.sslclientkey = kwargs.get("sslclientkey", None)
+
+    def __str__(self):
+        retval = KickstartCommand.__str__(self)
```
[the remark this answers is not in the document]
larskarlitski: I didn't do that, because `F27_Url.__str__(self)` adds a newline. I simply copied what the older classes were doing. I think removing the newline before adding the other options is a bit weird - what do you think?

pykickstart/version.py
```diff
@@ -87,10 +87,11 @@
 F27 = 25000
 F28 = 26000
 F29 = 27000
+F30 = 30000
 RHEL8 = 26100
 
 # This always points at the latest version and is the default.
-DEVEL = F29
+DEVEL = F30
```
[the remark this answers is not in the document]
larskarlitski: Ugh, I simply forgot to `git add` it. Sorry :smirk:

pykickstart/commands/repo.py
larskarlitski: Indeed. Thanks.

larskarlitski:
Update to add tests and address @poncovka's points. Thanks for the review!

> Please, don't forget to run make test and make check. The automated tests work only partially.

Done, and works on Python previous to 3.7. I've filed  https://github.com/dcantrell/pykickstart/pull/251 to make it work on 3.7 as well.

pykickstart/commands/url.py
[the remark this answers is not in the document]
poncovka: Yes, it is weird, but it is a common practice in pykickstart. Look at other commands, for example:
https://github.com/dcantrell/pykickstart/blob/master/pykickstart/commands/clearpart.py#L162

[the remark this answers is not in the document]
larskarlitski: Interesting. Thanks for the pointer. I've pushed a change.

pykickstart/commands/repo.py
```diff
@@ -423,5 +447,22 @@ def _getParser(self):
...
 
+class F30_Repo(F27_Repo):
+    removedKeywords = F27_Repo.removedKeywords
+    removedAttrs = F27_Repo.removedAttrs
+
+    def _getParser(self):
+        op = F27_Repo._getParser(self)
+        op.add_argument("--sslcacert", version=F30, help="""
+                        Path to the directory or file containing the
+                        certificate authorities to verify SSL certificates.""")
```
M4rtinK: It might be good to more clearly specify which file formats & directory layouts are supported for this.

pykickstart/commands/repo.py
```diff
@@ -423,5 +447,22 @@ def _getParser(self):
...
+        op.add_argument("--sslclientcert", version=F30, help="""
+                        Path to the SSL client certificate used to connect to
```
M4rtinK: Here it also might make sense to be more specific what certificate formats are supported & possibly how the path could look like ("file://foo, /foo, foo ?) etc.

pykickstart/commands/repo.py
larskarlitski: Hm good point. I took this documentation from `man dnf.conf`. It sadly doesn't specify further what is allowed.

Do you think it's enough to point to "what dnf accepts", or should we be more strict?

larskarlitski:
I've added a commit to add those options to the `RHEL8Handler` as well to make it usable on rhel 8 beta. Is this the right way to go about that?

poncovka:
> I've added a commit to add those options to the `RHEL8Handler` as well to make it usable on rhel 8 beta. Is this the right way to go about that?

It looks correct, but you should also open a pull request for `rhel8-branch`.

pykickstart/commands/repo.py
dcantrell: I agree with Martin here, the pykickstart help string should explain more.  I do not think it is sufficient to say just look at the man page for dnf.conf because, historically, the yum and dnf documentation has also been lacking.  Plus it's kind of annoying when you're reading some docs and are told to go jump in to another doc for one specific thing.  It's ok to say "see also dnf.conf(5)" but I would prefer that the pykickstart documentation be explicit.  This information makes its way in to both the Fedora and RHEL install guides as well.

pykickstart/commands/repo.py
dcantrell: Agreed with Martin here.  This help string needs to be more explicit.

dcantrell:
General comment and concern.... are these kickstart options deemed ok to include in upstream pykickstart?  For a long time, the code surrounding subscription-manager and certificates were considered specific to RHEL.  If that has changed, I am ok with it, but I would like to hear from product management that we are ok including this functionality in pykickstart in the upstream code and, subsequently, releases that ship in Fedora--even if the functionality in Fedora is not hooked up to anything.  If we can't do that and we must keep it explicitly RHEL-only, we'll need to figure out another way to incorporate these changes.

larskarlitski:
> General comment and concern.... are these kickstart options deemed ok to include in upstream pykickstart? [...]

Good point. However, this is not inherently specific to subscriptions. These are general options for rpm repositories and are [documented upstream](https://dnf.readthedocs.io/en/latest/conf_ref.html). You're right though, client certificates are probably only used by RHEL subscriptions at the moment. 

> It looks correct, but you should also open a pull request for `rhel8-branch`.

Ok, I'll do that once everyone's ok with these changes.

pykickstart/commands/repo.py
```diff
@@ -168,6 +168,30 @@ def _getArgsAsStr(self):
...
+        retval = F27_RepoData._getArgsAsStr(self)
+
+        if self.sslcacert:
+            retval += " --sslcacert=\"%s\"" % self.sslcacert
+
+        if self.sslclientcert:
+            retval += " --sslclientcert=\"%s\"" % self.sslclientcert
+
+        if self.sslclientkey:
+            retval += " --sslclientkey=\"%s\"" % self.sslclientkey
```
atodorov: Are all of these 3 required ?

larskarlitski: Yes :)

pykickstart/handlers/f30.py
```diff
@@ -0,0 +1,123 @@
+#
+# Vendula Poncova <vponcova@redhat.com>
```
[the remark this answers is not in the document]
larskarlitski: Feels weird on a file that I mostly copied, but the previous ones seem to be similar.

pykickstart/handlers/f30.py
```diff
@@ -0,0 +1,123 @@
...
+#
+# Copyright 2018 Red Hat, Inc.
```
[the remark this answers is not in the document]
larskarlitski: Indeed. Happy new year! :balloon:

larskarlitski:
> Except these two nitpicks looks good to me.

Fixed and rebased to master. Thanks for the review!

larskarlitski:
@M4rtinK, @dcantrell I've pushed a slight change to the descriptions. As I said, `dnf` doesn't specify this further. Reading the code, I found out that it delegates to libcurl's [CURLOPT_CAINFO](https://curl.haxx.se/libcurl/c/CURLOPT_CAINFO.html), [CURLOPT_SSLCERT](https://curl.haxx.se/libcurl/c/CURLOPT_SSLCERT.html) and [CURLOPT_SSLKEY](https://curl.haxx.se/libcurl/c/CURLOPT_SSLKEY.html). All of them are paths (not urls), as documented.

pykickstart/handlers/f30.py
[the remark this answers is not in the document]
jkonecny12: Yeah personally I would rather stop using this at all. You are able to find much more from the git so I don't see a point in it. @dcantrell what do you think?

larskarlitski:
Rebased. Tests are now green.

dcantrell:
This looks reasonable now, though I wish the documentation was more verbose.  I say that based on past experiences where users looked to the install guide for information even if a kickstart command was effectively a pass through to another command or library.  I suppose that can be expanded later if need be.

Thanks for the work on this.

PR #251, opened 2019-01-03
parser: use collections.abc to import Iterator

larskarlitski:
`collections.abc` was introduced in python 3.3. Importing these classes
from `collections` directly will be deprecated in 3.8 and already makes
the linting fail.

Note: we should move this into a `try` / `except ImportError` block if we care about python < 3.3. Do we?

### DIFF
--- a/pykickstart/parser.py
+++ b/pykickstart/parser.py
@@ -33,7 +33,7 @@ This module exports several important classes:
 
 from __future__ import print_function
 
-from collections import Iterator
+from collections.abc import Iterator
 import os
 import six
 import shlex


larskarlitski:
Another fix: sphinx 1.8 changed an import path. I had an earlier version installed in my environment.

dcantrell:
There's kind of a lot going on in this patch set.  Is it all related to the collections.abc change as stated in the summary or is it an "oh, and also" type of thing?  I know it's really easy to fall down that hole, but I would like pull requests to be logically separated and in order.  So if the Makefile change needs to happen first, make that a PR.  The changes in these files all look fine to me, but I would prefer them to be logically separated out and tied to different PRs.  Thanks.

larskarlitski:
Indeed, sorry about that. I've used the opportunity to also fix CI, but that's clearly unrelated. I've split those two commits into https://github.com/dcantrell/pykickstart/pull/252.

Note that CI will be red again for this. Let's land the other one first and I'll rebase this.

dcantrell:
Thanks, I'll look over the other one.

larskarlitski:
Rebased. Tests are green now.

PR #253, opened 2019-01-23
Fix backward compatibility for python2

jkonecny12:
The `collections.abc` python module was introduced from python 3.3 but pykickstart still supports python 2 builds.

Introduced by PR https://github.com/dcantrell/pykickstart/pull/251.

@dcantrell I have found that thanks to our daily builds. You have now two options.

1) Merge this or similar patch (blue pill)
2) Drop python2 package from pykickstart (red pill)

The choice is your. Will you follow the white rabbit down to the rabbit hole?

### DIFF
--- a/pykickstart/parser.py
+++ b/pykickstart/parser.py
@@ -33,7 +33,11 @@ This module exports several important classes:
 
 from __future__ import print_function
 
-from collections.abc import Iterator
+try:
+    from collections.abc import Iterator
+except ImportError:  # python2 compatibility
+    from collections import Iterator
+
 import os
 import six
 import shlex


larskarlitski:
Heh, I asked about this on #251, but didn’t get a response. Thanks!

jkonecny12:
Yeah, I saw that but I didn't realize that there is still python2 package of pykickastart.

jkonecny12:
So the build seems to be working correctly even without the patch.
The original exception was raised by building documentation which was changed to be done by python3 in commit:

6f21937f9f30a934013869a27de6b87120213410

However, I wonder how is it possible that the package is built when there is no such a thing like `collections.abc` in python 2. My personal guess is that the tests are not run and build process don't invoke the code in a process.

If I'm correct than we are building python2-pykickstart package in Rawhide which just doesn't work but we don't know that because no-one using it. I try to test my assumptions.

jkonecny12:
As I expected:
```
<mock-chroot> sh-4.4# python2
Python 2.7.15 (default, Jan 14 2019, 22:22:39) 
[GCC 8.2.1 20190109 (Red Hat 8.2.1-7)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> from pykickstart import parser
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python2.7/site-packages/pykickstart/parser.py", line 36, in <module>
    from collections.abc import Iterator
ImportError: No module named abc
```

@dcantrell the new version release you did is with broken package. This should be fixed soon.

M4rtinK:
Fedora 30 branching is 2019-02-19 (https://fedoraproject.org/wiki/Releases/30/Schedule), so I guess if Python 2 support in Pykickstart is to be dropped, it would be good to drop the Python 3 sub-packages till then (and possibly drop a note about that on fedora-devel ML).

dcantrell:
Thanks, everyone.  I will check on system-config-kickstart and see what options exist there.

PR #256, opened 2019-02-14
Test on recent Python versions on Travis

M4rtinK:
At the moment 3.6 is in RHEL 8 and 3.7 in Fedora,
so I think it makes sense to run the Pykickstart
test suite on those in Travis as well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,8 @@
 language: python
 python:
   - "3.5"
+  - "3.6"
+  - "3.7"
 
 install:
   - pip install --upgrade pip setuptools


poncovka:
Python 3.7 doesn't seem to be available.

```
3.7 is not installed; attempting download
Downloading archive: https://s3.amazonaws.com/travis-python-archives/binaries/ubuntu/14.04/x86_64/python-3.7.tar.bz2
$ curl -sSf -o python-3.7.tar.bz2 ${archive_url}
curl: (22) The requested URL returned error: 403 Forbidden
Unable to download 3.7 archive. The archive may not exist. Please consider a different version.
```

PR #258, opened 2019-02-19
Fix snapshot documentation

jkonecny12:
The snapshot origin was missed in the documentation which was done mostly thanks to the bad parsing.


pykickstart/commands/snapshot.py
```diff
@@ -110,17 +113,9 @@ def _getParser(self):
     def parse(self, args):
         (ns, extra) = self.op.parse_known_args(args=args, lineno=self.lineno)
```
poncovka: You can call `parse_args` now.

It will probably fix the pylint warning:
`W0612(unused-variable):114,13: F26_Snapshot.parse: Unused variable 'extra'`

jkonecny12: Fixed.

jkonecny12:
*UPDATED*

PR #259, opened 2019-03-04
Configuration file .packit.yaml

phracek:
Signed-off-by: Petr "Stone" Hracek <phracek@redhat.com>

This PR adds support for packit https://github.com/packit-service/packit, which syncs upstream changes into downstream automatically. See docu on packit side https://github.com/packit-service/packit/blob/master/README.md.

PR #262, opened 2019-03-21
Add pykickstart.spec.in for packit

phracek:
Signed-off-by: Petr "Stone" Hracek <phracek@redhat.com>

This pull request is to follow up #259 .

The upstream repository contains temporary `pykickstart.spec.in` the file which is generated by Makefile `make pykickstart.spec`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -109,9 +109,8 @@ bumpver: docs
 	make -C po pykickstart.pot ; \
 	zanata push $(ZANATA_PUSH_ARGS)
 
-pykickstart.spec:
-	cp pykickstart.spec.in pykickstart.spec
-	sed -i "s/%%VERSION%%/$(VERSION)/" pykickstart.spec
+pykickstart.spec: pykickstart.spec.in
+	sed -e "s/%%VERSION%%/$(VERSION)/" < $< > $@
 
 scratch-bumpver: docs
 	@NEWSUBVER=$$((`echo $(VERSION) |cut -d . -f 2` + 1)) ; \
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -109,6 +109,10 @@ bumpver: docs
 	make -C po pykickstart.pot ; \
 	zanata push $(ZANATA_PUSH_ARGS)
 
+pykickstart.spec:
+	cp pykickstart.spec.in pykickstart.spec
+	sed -i "s/%%VERSION%%/$(VERSION)/" pykickstart.spec
```
dcantrell: Style concern here.  Can you do this in one line instead, like:

sed -e "s/%%VERSION%%/$(VERSION)/" < $< > $@

Changing the target so it lists pykickstart.spec.in as a requirement

Makefile
```diff
@@ -109,6 +109,10 @@ bumpver: docs
 	make -C po pykickstart.pot ; \
 	zanata push $(ZANATA_PUSH_ARGS)
 
+pykickstart.spec:
```
dcantrell: Change to:

pykickstart.spec: pykickstart.spec.in

Makefile
phracek: This can be solved also by the command `sed "s/%%VERSION%%/$(VERSION)/" $< > $@`.
But I will use your approach.

dcantrell:
Thanks for the update.  I'm going to merge this now, but there's still a bit I want to do for the spec file before it's ready to generate packages for submission via packit.

PR #263, opened 2019-03-28
Add options for ssl certs on url and repo commands

larskarlitski:
This adds --sslclientcert, --sslclientkey, and --sslcacert options to
the `url` and `repo` commands. They mirror the options in dnf.conf.


poncovka:
Hi, usually we don't add Fedora handlers to RHEL branches and we use the RHEL version in names of new commands and data classes.

larskarlitski:
> Hi, usually we don't add Fedora handlers to RHEL branches and we use the RHEL version in names of new commands and data classes.

Hm, I simply cherry-picked the commit from `master`. The changes that I proposed for lorax and anaconda depend on these classes. Should their RHEL branches use these classes instead? Or do we have both?

poncovka:
@larskarlitski, yes, the RHEL branches should use RHEL classes. At least, it worked this way on RHEL 7. It makes cherry-picking more difficult, but it is easier to see what has changed in the kickstart commands after branching from Fedora. It is not an ideal solution, but we don't have a better one right now.

stefwalter:
Ping @dcantrell anything else preventing this from being merged?

dcantrell:
Hey everyone.  I'm merging this to rhel8-branch and will be doing a new release for that branch.  For urgent things, please find me in IRC or send me an email in the future.

dcantrell:
Also, Travis-CI is dead right now and I will be working on that after I do this release.  Just so everyone knows.

PR #264, opened 2019-04-01
Make %packages --default attribute sane

jkonecny12:
The ``%packages --default`` will now force default environment. User can't change the environment even when he thinks he did. As a bonus Anaconda will use the user-set environment everywhere but not for installation.

All groups and packages written in the ``%packages`` section are installed correctly but they are not propagated to the output kickstart file.

In this commit I'm changing the output section generator to print all the groups and packages to the output kickstart but avoiding environment, which will be forced and can't be changed. Anaconda may change this behavior in the future but in that case the `--default` is not used and will be disabled.

Also change documentation to mention that only environment is set nothing else.

### DIFF
--- a/pykickstart/parser.py
+++ b/pykickstart/parser.py
@@ -371,36 +371,8 @@ class Packages(KickstartObject):
 
     def __str__(self):
         """Return a string formatted for output to a kickstart file."""
-        pkgs = ""
-
-        if not self.default:
-            if self.environment:
-                pkgs += "@^%s\n" % self.environment
-
-            grps = self.groupList
-            grps.sort()
-            for grp in grps:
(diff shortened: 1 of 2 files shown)


jkonecny12:
**UPDATED:**
* I've broke correct output section generator based on an empty packages section and seen argument. Tests have found that luckily.

dcantrell:
Patch looks good.  Tests failed due to a change I had to make to the .travis.yml file on master.  They should be running fine again.  Can you rebase your branch from master and resubmit this so it can run the test suite again?  I don't think I can do that easily from here.

jkonecny12:
**UPDATED:**
* Rebased

PR #265, opened 2019-04-11
packit.yml: remove unused content

TomasTomecek:
We are working on a next release of packit where we'll finally implement jobs.
Our current plan is to change the schema, which means that future version of
packit would fail parsing this config file.

This PR therefore removes content from packit.yml which packit does not even
process now.


dcantrell:
Thank you for the update, merging.

PR #273, opened 2019-07-26
Fix outdate information about lang in the docs

jkonecny12:
The lang command documentation is pointing on ``/usr/share/system-config-language/locale-list`` from a ``system-config-languages`` package package but that is AFAIK deprecated by systemd ``localectl`` command.

### DIFF
--- a/pykickstart/commands/lang.py
+++ b/pykickstart/commands/lang.py
@@ -55,9 +55,8 @@ class FC3_Lang(KickstartCommand):
             will continue in English though the running system will have the
             specified langauge by default.
 
-            The file ``/usr/share/system-config-language/locale-list`` provides a
-            list the valid language codes in the first column of each line and
-            is part of the system-config-languages package.""", version=FC3)
+            You can list the valid language codes by calling
+            ``localectl list-locales``.""", version=FC3)
         op.add_argument("lang", metavar="<lang>", nargs=1, version=FC3,
                         help="Language ID.")
         return op


pykickstart/commands/lang.py
```diff
@@ -55,9 +55,8 @@ def _getParser(self):
             will continue in English though the running system will have the
             specified langauge by default.
 
-            The file ``/usr/share/system-config-language/locale-list`` provides a
-            list the valid language codes in the first column of each line and
-            is part of the system-config-languages package.""", version=FC3)
+            You can list the valid language codes by calling
+            ``localectl list-locales``.""", version=FC3)
```
poncovka: We don't support all of them. It should be  `localectl list-locales | grep _`. Also `locale -a | grep _` works.

jkonecny12: Heh, True I've missed those "special" locales listed there. I wonder if there is not a better way to get the list. I'll investigate that a little bit.

poncovka: I was already looking into this for the RHEL documentation, but I couldn't find a better way. We are limited by the langtable, but the langtable doesn't support these special cases and doesn't provide a list of supported locales, so there is nothing else we can do.

jkonecny12: I want to ask creator of the library we are using where and how he is taking data. Just in case he would have better way to do that.

mike-fabian: > We don't support all of them. It should be `localectl list-locales | grep _`. Also `locale -a | grep _` works.

"localectl  list-locales" seems to do exactly the same as "locale -a". If I install extra locales by installing extra glibc locale packages, both list the additionally installed locales. I guess "localectl list-locales" just calls "locale -a". 

Most of the locale names listed by "locale -a" which do not contain an _ are legacy aliases.
But a few of them are  real, useful locales. These ones:

C
C.utf8
POSIX
eo.UTF-8  <- this is Esperanto

mike-fabian: > We don't support all of them.

I wonder  what you mean by "support".

Supported by glibc? That would be the output of "locale -a" or "localectl list-locales", but it depends on which glibc subpackages are installed. 

glibc-all-langpacks-2.29-15.fc30.x86_64 contains all of them. 

But one can also install only a few packages like 

glibc-langpack-en-2.29-15.fc30.x86_64
glibc-langpack-es-2.29-15.fc30.x86_64
...

then "locale -a" will list far fewer locales. 

Or do we mean by "supported" the languages where we have translations for? Or maybe the languages where anaconda has translations for?

At the start of the installation, anaconda shows a language selection page. The list of languages shown there is far smaller than the list of locales supported by glibc. 

As far as  I know this list is generated from the list of language where anaconda has translations for.

mike-fabian: > I was already looking into this for the RHEL documentation, but I couldn't find a better way. We are limited by the langtable, but the langtable doesn't support these special cases and doesn't provide a list of supported locales, so there is nothing else we can do.

I think the legacy aliases like "czech", "deutsch", etc... should be ignored and not used anymore.

eo.utf8 is a useful locale though and it is listed in langtable:

```
$ python3 
Python 3.7.4 (default, Jul  9 2019, 16:32:37) 
[GCC 9.1.1 20190503 (Red Hat 9.1.1-1)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import langtable
>>> langtable.list_locales(languageId='eo')
['eo.UTF-8']
```

mike-fabian: > I was already looking into this for the RHEL documentation, but I couldn't find a better way. We are limited by the langtable, but the langtable doesn't support these special cases and doesn't provide a list of supported locales, so there is nothing else we can do.

langtable tries to be as complete as possible in storing information about languages and territories. So langtable even includes languages where we don’t even have a locale in glibc.
For example:

```
$ python3 
Python 3.7.4 (default, Jul  9 2019, 16:32:37) 
[GCC 9.1.1 20190503 (Red Hat 9.1.1-1)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import langtable
>>> langtable.list_locales(languageId='av')
[]
>>> langtable.language_name(languageId='av')
'авар мацӀ'
>>> langtable.language_name(languageId='av', languageIdQuery='en')
'Avaric'
>>> 
```

So glibc currently has no locale available for “Avaric”. One could still store translations for “Avaric”
in

/usr/share/locale/av/LC_MESSAGES/

and set LANGUAGE="av" to get gettext translations for such a language which doesn’t yet have a glibc locale.

Currently, langtable does not have a function to return a list of locales supported by glibc.

Should I add such a function? Would that be helpful?

Currently, the list_locales() function returns nothing if called with an empty languageId or with no parameters at all:

```
>>> langtable.list_locales(languageId='')
[]
>>> langtable.list_locales()
[]
>>>
```

I could change that to list all available locales, i.e. all locales supported by glibc (without third party extras, users can add locales to glibc using localedef). Would that be helpful? Should I add something like that?

jkonecny12: Wow thanks a lot for your input @mike-fabian!

The problem is that Anaconda use other checks than you have. We are using this regex to parse the input for langtable or other usage:

https://github.com/rhinstaller/anaconda/blob/master/pyanaconda/localization.py#L38

So basically we won't accept anything what don't have `_`. Not sure if that is a good way to go.


About your proposed solution. It won't help us in the code, we want that for users to know what they should specify for the installation to work. So we might have think about this as a binary what user will call to get list of supported languages but I would more prefer to stay as close as possible to the running system.

What do you think?

mike-fabian: I still don’t understand completely what is needed. Users who are going to do an installation and want to do what  languages are supported? Don’t they see that in the lists displayed on the first page shown by anaconda right at the start of the installation?

In the running system, "locale -a" shows the locales which are currently installed. Which could be all supported locales  if the glibc-all-langpacks package is installed. Or it could be a subset of these if only some glibc-langpack-?? packages are installed.

jkonecny12: > I still don’t understand completely what is needed. Users who are going to do an installation and want to do what languages are supported? Don’t they see that in the lists displayed on the first page shown by anaconda right at the start of the installation?

The problem is that user are able to decide the language based on the kernel boot parameter `inst.lang` and `inst.keyboard`. However, to be able to do that, they have to know the code for their language. We are trying to find out the best way to give them the list of what they can use.

> In the running system, "locale -a" shows the locales which are currently installed. Which could be all supported locales if the glibc-all-langpacks package is installed. Or it could be a subset of these if only some glibc-langpack-?? packages are installed.

So that means that user have to install all the `glibc-langpack-*` to get the correct list of what is supported in Anaconda? Of course with the good filters of the list (mean the ``grep |``).

jkonecny12: @mike-fabian ping, are you fine with merging the ``localectl list-locales | grep _`` solution?

mike-fabian: Yes, I think that is OK, although you miss eo.UTF-8 and  C.UTF-8. These are currently the only useful UTF-8 locales which don’t  have an _ in their locale name.

jkonecny12:
> Is there any way Anaconda can change to accept the full list? The (small) issue I have with changing this is that the `_` limitation is specific to the Anaconda use of pykickstart, and not a limitation of pykickstart itself.

Good question but someone need to investigate what will happen...

jkonecny12:
So this is a question on @mike-fabian really. Right now we are blocked by how langtable is taking parameters. What is right now happening is this:

1) Get the lang code from a user.
2) Take the lang code and parse that:
https://github.com/rhinstaller/anaconda/blob/master/pyanaconda/localization.py#L54
3) Use this to get information from a langtable:
e.g. https://github.com/rhinstaller/anaconda/blob/master/pyanaconda/localization.py#L308


What would be great if similar parsing (2) would be part of the langtable including the legacy variants and translated to something we can then use again in the langtable. In other words to support translation of the legacy variants to a new solution. @mike-fabian will that work for langtable, is it doable?

mike-fabian:
> So this is a question on @mike-fabian really. Right now we are blocked by how langtable is taking parameters. What is right now happening is this:
> 
>     1. Get the lang code from a user.
> 
>     2. Take the lang code and parse that:
>      def parse_langcode(langcode):

[Here](https://github.com/rhinstaller/anaconda/blob/master/pyanaconda/localization.py#L54)
I see a typo:

```
def parse_langcode(langcode):
    """
    For a given langcode (e.g. 'SR_RS.UTF-8@latin') returns a dictionary
    with the following keys and example values:
    'language' : 'SR'
    'territory' : 'RS'
    'encoding' : 'UTF-8'
    'script' : 'latin'
    or None if the given string doesn't match the LANGCODE_RE.
    """  
```

language should be lowercase, i.e. not SR_RS.UTF-8@latin but sr_RS.UTF-8@latin.

mike-fabian:
> So this is a question on @mike-fabian really. Right now we are blocked by how langtable is taking parameters. What is right now happening is this:
> 
>     1. Get the lang code from a user.
> 
>     2. Take the lang code and parse that:
>        https://github.com/rhinstaller/anaconda/blob/master/pyanaconda/localization.py#L54
> 
>     3. Use this to get information from a langtable:
>        e.g. https://github.com/rhinstaller/anaconda/blob/master/pyanaconda/localization.py#L308
> 
> 
> What would be great if similar parsing (2) would be part of the langtable including the legacy variants and translated to something we can then use again in the langtable. In other words to support translation of the legacy variants to a new solution. @mike-fabian will that work for langtable, is it doable?

Yes.
And at least partly it already works like that:

```
$ python3
Python 3.7.4 (default, Jul  9 2019, 16:32:37) 
[GCC 9.1.1 20190503 (Red Hat 9.1.1-1)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import langtable
>>> langtable.language_name(languageId="sr_RS.UTF-8@latin")
'Srpski (Latinica) (Srbija)'
>>> 
```

I am checking now whether this works for *all* (UTF-8) locale names in glibc,
if not I’ll try to fix it ... ⏳

jkonecny12:
Nice thanks a lot for looking into this!

mike-fabian:
> Nice thanks a lot for looking into this!

https://github.com/mike-fabian/langtable/releases/tag/0.0.46

This new release fixes only some minor problems, for example
language_name(languageId="quz") returned an empty string because
there was no data for the language endonym for the  language "quz" (Cusco Quechua)
in langtable. langtable did already parse languageId into parts when a full locale  name
was given for a long time. So something like language_name(languageId="sr_RS.UTF-8@latin")
always worked. The new  release only makes  sure that language_name() returns something
for *all* UTF-8 locales currently in glibc. For most of these it already worked, but glibc has locales
for some really unusual language where not even CLDR has any information. So I added 
language endonyms and some territory translations for all these unusual languages as well.

And  I added some documentation in  the comment near the top of langtable.py. 
See:

https://github.com/mike-fabian/langtable/blob/0.0.46/langtable/langtable.py#L38

VladimirSlavik:
@mike-fabian I have taken over this problem area from @jkonecny12 somewhat and started by replacing our custom parsing with langtable wherever possible. So far I have identified the following reasons why langtable as is can't satisfy all our needs:

1. No way to **get just language ID** - as in, we'd like to ask about `SR_RS.UTF-8@latin` and get `SR`. The previous discussion centered around making your internal parsing available, but I am now almost certain that languageId is the only component of locale that we really consistently need throughout the whole works.

    I am not sure what you see as more desirable on your end? If making only one component available is okay, I'd go for either adding something like `languadeId_from_locale()` or extending `languageId()` to ingest locale somehow. Bear in mind I have not looked at langtable, so it's just an opening of a discussion.

2. You do not handle **encoding** at all. I think unlike the previous point this would effectively need a complete rewrite of langtable, so I consider that a no-go. Is that true?

   This leaves us with a few places where we must retain our custom parsing. Mostly these are places that determine if things match:
    - [`langcode_matches_locale()`](https://github.com/rhinstaller/anaconda/blob/master/pyanaconda/localization.py#L111)
    - [`find_best_locale_match()`](https://github.com/rhinstaller/anaconda/blob/master/pyanaconda/localization.py#L143)

mike-fabian:
>     2. You do not handle **encoding** at all. I think unlike the previous point this would effectively need a complete rewrite of langtable, so I consider that a no-go. Is that true?

Yes, I think I do not handle encoding, but I am not sure what exactly
you are asking. I do parse locale names like de_DE.ISO-8859-15@euro
into its components but I just throw away the encoding part.

What do you mean by handling encoding? What kind of handling would you
want or would be useful to you?

mike-fabian:
>     1. No way to **get just language ID** - as in, we'd like to ask about `SR_RS.UTF-8@latin` and get `SR`. The previous discussion centered around making your internal parsing available, but I am now almost certain that languageId is the only component of locale that we really consistently need throughout the whole works.

The internal function  _parse_and_split_languageId() can do something like this:

```
    >>> _parse_and_split_languageId(languageId='sr_RS.UTF-8@latin')
    ('sr', 'Latn', 'RS')
```

You can't use it because it is internal, but if I add something like that to the public API, would it do what you need?

It returns a tuple with three strings, the first is the ISO  language code,  the second  the ISO script code and  the last the ISO country code.

>        I am not sure what you see as more desirable on your end? If making only one component available is okay, I'd go for either adding something like `languadeId_from_locale()` or extending `languageId()` to ingest locale somehow. Bear in mind I have not looked at langtable, so it's just an opening of a discussion.

The public functions in langtable usually have the three parameters languageId, scriptId, and territoryId, for example:

```
def list_locales(concise=True, show_weights=False, languageId = None, scriptId = None, territoryId = None):
```

You do not have to use the scriptId and territoryId parameters, you can stick all into the languageId parameter if  you prefer.
For example:

```
$ python3
Python 3.7.5 (default, Oct 17 2019, 12:16:48) 
[GCC 9.2.1 20190827 (Red Hat 9.2.1-1)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import langtable
>>> langtable.list_locales(languageId='sr', scriptId='Latn', territoryId='RS')
['sr_RS.UTF-8@latin']
>>> langtable.list_locales(languageId='sr_Latn_RS')
['sr_RS.UTF-8@latin']
>>> langtable.list_locales(languageId='sr_RS.UTF-8@latin')
['sr_RS.UTF-8@latin']
```

You can always stick a full glibc locale name into the languageId  parameter and  leave out the scriptId and territoryId parameters.

If you supply the same information both in the languageId paramter and in some other parameter like for example territoryId,
the languageId parameter wins. For example:

```
>>> langtable.list_locales(languageId='sr_RS.UTF-8@latin', territoryId='ME')
['sr_RS.UTF-8@latin']
```

ignores the 'ME' territory because languageId already specifies 'RS'.

```
>>> langtable.list_locales(languageId='sr', territoryId='ME')
['sr_ME.UTF-8']
>>> langtable.list_locales(languageId='sr_ME.UTF-8')
['sr_ME.UTF-8']
>>> 
```

Does that do what you need?

mike-fabian:
I wrote

> You can always stick a full glibc locale name into the languageId parameter and leave out the scriptId > and territoryId parameters.

As langtable ignores the encoding, you can leave out the encoding of the glibc locale name and get the same result.

mike-fabian:
```
def langcode_matches_locale(langcode, locale):
    """
    Function that tells if the given langcode matches the given locale. I.e. if
    all parts of appearing in the langcode (language, territory, script and
    encoding) are the same as the matching parts of the locale.
    :param langcode: a langcode (e.g. en, en_US, en_US@latin, etc.)
    :type langcode: str
    :param locale: a valid locale (e.g. en_US.UTF-8 or sr_RS.UTF-8@latin, etc.)
    :type locale: str
    :return: whether the given langcode matches the given locale or not
    :rtype: bool
    """
 ```

As the my internal function _parse_and_split_languageId() ignores the encoding
and returns the other three parts, you could rewrite that function like this
if _parse_and_split_languageId() were a public function:

```
def langcode_matches_locale(langcode, locale):
    return _parse_and_split_languageId(langcode) == _parse_and_split_languageId(locale)
```

The following examples show that this would work:

```
    >>> _parse_and_split_languageId(languageId='sr_RS.UTF-8@latin') == _parse_and_split_languageId(languageId='sr_RS@latin')
    True

    >>> _parse_and_split_languageId(languageId='sr_RS.UTF-8@latin') == _parse_and_split_languageId(languageId='sr_Latn_RS')
    True

    >>> _parse_and_split_languageId(languageId='sr_RS.UTF-8@latin') == _parse_and_split_languageId(languageId='sr_RS.UTF-8')
    False

    >>> _parse_and_split_languageId(languageId='sr_RS.UTF-8@latin') == _parse_and_split_languageId(languageId='sr_RS')
    False
```

It does not assume a default script though if it the script is not specified,
for example here you get False:

```
    >>> _parse_and_split_languageId(languageId='sr_RS.UTF-8@latin') == _parse_and_split_languageId(languageId='sr_RS')
    False
```

because the Latin script is specified in 'sr_RS.UTF-8@latin" but no script is specified in 'sr_RS:

```
    >>> _parse_and_split_languageId(languageId='sr_RS')
    ('sr', None, 'RS')
    >>> _parse_and_split_languageId(languageId='sr_RS.UTF-8@latin')
    ('sr', 'Latn', 'RS')
```

The glibc locale sr_RS.UTF-8 uses Cyrillic script. and comparing sr_RS.UTF-8 with sr_RS would give True. But comparing sr_RS.UTF-8 with sr_Cyrl_RS would give false because

```
    >>> _parse_and_split_languageId(languageId='sr_RS.UTF-8')
    ('sr', None, 'RS')
```

It returns None as the script because it only parse the string 'sr_RS.UTF-8' and this string does not say that it this locale uses Cyrillic script. So the script returned is always None if it is not found in the string.

**Comment by VladimirSlavik** [comment]
Hi @mike-fabian , looking at your examples, that would go a long way towards outmoding our messy parsing :) So, if you are comfortable making the parsing function public, then yes please?

**Comment by mike-fabian** [comment]
> Hi @mike-fabian , looking at your examples, that would go a long way towards outmoding our messy parsing :) So, if you are comfortable making the parsing function public, then yes please?

OK, I’ll do that. Would you prefer if it returns the encoding part as well?
And, would you prefer the return value to be a dictionary (like in your parsing function in anaconda) or a tuple (as it is  now in the parsing function in  langtable)?

**Comment by VladimirSlavik** [comment]
If you can add encoding, that would be fantastic! As to dictionary/tuple, I'd say do what makes sense to you or suits you better. That said, if you asked me, I think a dictionary (or even object?) is more clean than hiding the meaning in 0,1,2...

**Comment by mike-fabian** [comment]
OK, then I'll use a dictionary as the return value.

**Comment by poncovka** [comment]
Hi, I would prefer a namedtuple, so we don't have to use indexing at all.

**Comment by mike-fabian** [comment]
> Hi, I would prefer a namedtuple, so we don't have to use indexing at all.

I didn’t know namedtuples, thank you for suggesting that, it looks very nice for that purpose, I’ll use a namedtuple.

**Comment by mike-fabian** [comment]
How about this?:

https://github.com/mike-fabian/langtable/blob/release-0.0.49/langtable/langtable.py#L974

**Comment by VladimirSlavik** [comment]
That's perfect, Mike! Thank you!

**Comment by mike-fabian** [comment]
I released  it:
https://github.com/mike-fabian/langtable/releases/tag/0.0.49
https://bodhi.fedoraproject.org/updates/FEDORA-2019-38792d4e43
https://bodhi.fedoraproject.org/updates/FEDORA-2019-0fbca98a89

**Comment by VladimirSlavik** [comment]
Thank you Mike!

**Comment by bcl** [comment]
How does this effect pykickstart? Do the docs still need to be changed and if so, is this PR now an accurate representation of them? :)

**Comment by VladimirSlavik** [comment]
This PR is definitely not accurate any more, though to what degree is yet to be determined...

**Comment by VladimirSlavik** [comment]
Sorry to keep hijacking this, but as we already have a discussion going... @mike-fabian, I have some more questions unfortunately.

First, the locales you mentioned earlier do not fit the pattern: `C` and `POSIX`. Is that something that could be solved on the level of langtable? I'm not even sure if these would be added to the data, or code as special case.

Secondly, `localectl list-locales` gives among others `fi_FI.iso885915@euro`. Turns out the script (modifier) is silently dropped unless it belongs to a rather short list of known values. Is this an intentional design? I would expect that to be treated as a opaque string, or validated against the full list linked in the comment. Given glibc and localectl offer this separately from the non-euro locale, I guess dropping this is significant? It won't propagate through the application (anaconda), so I wonder if we should then say such input value is supported or not...

Somewhat more generally, I am curious whether modifiers should take all scripts or not. The page linked in langtable sources has tons more, and the field is called script... Note I am not asking for this, just wondering. What significance does specifying a script have, and can it take effect? Probably not unless there is a matching locale definition file to be used by glibc?

**Comment by bcl** [comment]
> Sorry to keep hijacking this, but as we already have a discussion going...

No problem! I just wanted to be sure when things were all sorted out that we make the right changes to pykickstart (or no changes).

**Comment by mike-fabian** [comment]
> Sorry to keep hijacking this, but as we already have a discussion going... @mike-fabian, I have some more questions unfortunately.
> 
> First, the locales you mentioned earlier do not fit the pattern: `C` and `POSIX`. Is that something that could be solved on the level of langtable? I'm not even sure if these would be added to the data, or code as special case.

I could handle these in the parsing somehow.

I  wonder what parse_locale('C') or parse_locale('POSIX') should return ...

maybe language=en variant=POSIX  ???

No idea at the moment, what would be useful to you?

C and POSIX are a very special case.

> Secondly, `localectl list-locales` gives among others `fi_FI.iso885915@euro`. Turns out the script (modifier) is silently dropped unless it belongs to a rather short list of known values. Is this an intentional design? I would expect that to be treated as a opaque string, or validated against the full list linked in the comment. Given glibc and localectl offer this separately from the non-euro locale, I guess dropping this is significant? It won't propagate through the application (anaconda), so I wonder if we should then say such input value is supported or not...

@euro is not a script, it stands for "with euro currency", it is sort of a variant of the German locale with the euro.

```
$ LC_ALL=de_DE locale charmap
ISO-8859-1
$ LC_ALL=de_DE@euro  locale charmap
ISO-8859-15
```

The @euro variant uses ISO-8859-15 which contains the Euro symbol, just de_DE
uses Latin1 which does not have the Euro. Both is latin script of course.

de_DE.UTF-8 of course also has the Euro symbol. This is kind of legacy, not really
needed anymore when you have UTF-8.

But @euro is definitely not a script, it is more like a variant.

glibc has various stuff behind the @, some things are scripts, some are not.

For example, ca_ES@valencia is the variant of the Catalan langauge spoken in Valencia.
In CLDR/libicu syntax this is ca_ES_VALENCIA.

Sometimes the part behind the @ is a script in glibc, for example in

```
sr_RS
sr_RS.utf8
sr_RS.utf8@latin
sr_RS@latin
```

it means Latin script, sr_RS is using Cyrillic script, sr_RS@latin is using Latin script.
In CLDR/icu syntax this would be sr_Cyrl_RS and sr_Latn_RS.

glibc is not even consistent in using the same modifier always

```
tt_RU
tt_RU.utf8
tt_RU.utf8@iqtelif
tt_RU@iqtelif
```

tt_RU@iqtelif is using Latin script, tt_RU is using Cyrillic script.

> Somewhat more generally, I am curious whether modifiers should take all scripts or not. The page linked in langtable sources has tons more, and the field is called script... Note I am not asking for this, just wondering. What significance does specifying a script have, and can it take effect? Probably not unless there is a matching locale definition file to be used by glibc?

I am not sure I understand the question.

You cannot specify a script for glibc locale, i.e. you cannot just add @latin
to any glibc locale name unless such a locale exists.

**Comment by mike-fabian** [comment]
https://github.com/mike-fabian/langtable/releases/tag/0.0.50

I had to fix an exception and  as I was working on langtable anyway, I added
some parsing of of the special cases C, POSIX, en_US_POSIX, and C.UTF-8:

```
>>> parse_locale('en_US_POSIX')
Locale(language='en', script='', territory='US', variant='POSIX', encoding='')

>>> parse_locale('POSIX')
Locale(language='en', script='', territory='US', variant='POSIX', encoding='')

>>> parse_locale('C')
Locale(language='en', script='', territory='US', variant='POSIX', encoding='')

>>> parse_locale('C.UTF-8')
Locale(language='en', script='', territory='US', variant='POSIX', encoding='UTF-8')
```

If you have better ideas what the result should be, I could change it,
but at least this is better than nothing for the moment.

**Comment by VladimirSlavik** [comment]
Let's perhaps start with - "I don't know what I am doing". I hoped you would know, or have an opinion.

For the C/POSIX support - thank you, and I guess it's a possibility. I am currently defining "valid" locales as having a language, so this definitely works for me. My concern is that no other user has input here, except for me - and I don't trust myself enough with this...

Scripts - point taken. So that means langtable should handle locales that exist for real, and not let user construct hypothetical ones. Looking back, we left undefined if the parsing supposed to be reversible, and what class of inputs should it handle. I am inclined to believe that if real locales have @euro, and I will parse real locales with the function, then that should come through.

Actually, maybe the wrong part is me still following the expectation that Anaconda can validate a locale by seeing if it has a language, and just passing the ball onto langtable. I'll see if I can produce a laser-precise definition of what it is we do (and don't).

**Comment by mike-fabian** [comment]
> So that means langtable should handle locales that exist for real, and not let user construct hypothetical ones.

Yes, I think so.

I thought about  treating stuff in glibc which comes after @ but is not a script
as a variant, i.e. treating @euro as a variant, i.e.

```
>>> parse_locale('de_DE@euro')
Locale(language='de', script='', territory='DE', variant='EURO', encoding='')
```

(not implemented yet, just an idea).

See also

http://userguide.icu-project.org/locale

for the locale name syntax CLDR/icu use.

Examples:

```
en_IE_PREEURO

fr@collation=phonebook;calendar=islamic-civil

sr_Latn_RS_REVISED@currency=USD
```

So there can be many options after @ but the script is never after @,
if the script is there it is between the language and the territory
and it is a 4 letter ISO code for the script first letter upper case
following 3 letter lower case.

After the territory there maybe a variant. Like the "PREEURO"
and the "REVISED" in the examples. Note that these are just examples,
they don’t really exist either. If you look at  what currently
exists in CLDR/icu locales:

https://github.com/unicode-org/cldr/tree/master/common/main

You see that the only two currently existing locales with variants are
ca_ES_VALENCIA and en_US_POSIX.

VladimirSlavik:
So, I think I had it backwards. We use langtable primarily to handle locales internally when selecting them in GUI - that is, making them up so to say. That exactly matches the stated purpose of langtable.

However, to my surprise, when a locale comes from the outside as specified "manually" in kickstart or boot options, I don't see it being checked at all. Thus the parsing of `@euro` et al is really not a concern, at this moment. Sorry for false alarms.

In terms of parsing, Anaconda is now (selfishly) better off, as it does not miss some of the special cases. That then means improved world languages' coverage, as langtable knows Esperanto and Catalan in Valencia.

I'll look into this more, but for the moment I think I'll try wrapping things up on our end, so that this moves towards mainline, if not anything else.

mike-fabian:
> Thus the parsing of @euro et al is really not a concern, at this moment.

But if I parse everything after @ which is not a script as variant, that would be also OK for you, if I understand you right, it would not change anything for you so I could do that if I like.

VladimirSlavik:
Oh, it's absolutely good for me - having the information helps. I hope to explore this whole area of anaconda in yet more depth, so that I can eventually walk away knowing it is going to work...

jkonecny12:
I think this should be solved and not just closed. Because the current documentation is pointing to something which does not exists in current Fedora if I'm not mistaken.

bcl:
> I think this should be solved and not just closed. Because the current documentation is pointing to something which does not exists in current Fedora if I'm not mistaken.

This issue got pretty long-winded and off-track, but I kept it open because people were making progress.

If there is something in pykickstart that needs to be changed please open a new issue or PR that describes exactly what needs to be done.

jkonecny12:
OK, I will. Thanks!

PR #274, opened 2019-08-09
Make the module --disable option available on Fedora

M4rtinK:
The first commit adds the F31 handler, while the second subclasses the latest version of the module command and makes it available in Fedora 31+ via the F31 handler.


M4rtinK:
@dcantrell Looks like we are ready to go to have this merged & new release for Fedora created. Thanks in advance! :)

PR #275, opened 2019-08-16
Add --priority option to repo command

dwrobel:
This adds --priority=<priority> option to
    the `repo` command. It mirrors the option in dnf.conf.

    See also: https://dnf.readthedocs.io/en/latest/conf_ref.html#repo-priority-label

### DIFF
--- a/pykickstart/commands/repo.py
+++ b/pykickstart/commands/repo.py
@@ -177,6 +177,7 @@ class F30_RepoData(F27_RepoData):
         self.sslcacert = kwargs.get("sslcacert", None)
         self.sslclientcert = kwargs.get("sslclientcert", None)
         self.sslclientkey = kwargs.get("sslclientkey", None)
+        self.priority = kwargs.get("priority", None)
 
     def _getArgsAsStr(self):
         retval = F27_RepoData._getArgsAsStr(self)
@@ -189,6 +190,9 @@ class F30_RepoData(F27_RepoData):
 
         if self.sslclientkey:
             retval += " --sslclientkey=\"%s\"" % self.sslclientkey
+
+        if self.priority:
+            retval += " --priority=%s" % self.priority
 
         return retval
 
@@ -312,6 +316,14 @@ class F8_Repo(FC6_Repo):
 
     def _getParser(self):
         op = FC6_Repo._getParser(self)
+        op.add_argument("--priority", type=int, version=F30, help="""
+                        An integer value to assign a priority to this repository.
+                        The priority value of this repository, default is 99.
+                        If there is more than one candidate package for a
+                        particular operation, the one from a repo with the lowest
+                        priority value is picked, possibly despite being less
+                        convenient otherwise (e.g. by being a lower version).
+                        """)
         op.add_argument("--cost", type=int, version=F8, help="""
                         An integer value to assign a cost to this repository.
                         If multiple repositories provide the same packages,


dwrobel:
Basically it allows to use packages from a particular repository _(in a similar way as supported by `dnf`)_.

I tested it successfully using `appliance-creator` for building an image of [fedora-arm-server](https://pagure.io/fork/dwrobel/fedora-kickstarts/blob/fedora-arm-server-rpi4/f/fedora-arm-server.ks) for `RPi 3/4` with downstream [kernel](https://github.com/fedberry/kernel/commits/4.19.y), however it also requires changes in the [livecd-tools](https://github.com/livecd-tools/livecd-tools/pull/134) and the latest [libsolv](https://bugzilla.redhat.com/show_bug.cgi?id=1733582#c10).

jkonecny12:
Hello @dwrobel , what is the difference between ``--priority`` and ``--cost``? The both solving what repo should be preferably used.

Conan-Kudo:
@bcl @dwrobel Any progress here? It blocks my ability to merge this feature in livecd-tools...

dwrobel:
>Can someone explain why this is needed

It's allow you to install the package from a particular repository _(despite of its version, could be lower than in any other available repository)_. So, it's completely different to what the `cost` option can be used for.

bcl:
No, the documentation isn't that clear. And from looking at the code in dnf I'm less convinced that this is needed. As far as I can see priority is using in a tuple with cost to sort the repository, I don't see any other use of it. So as long as priority is the same for the repos, cost should have the same effect.

Conan-Kudo:
@bcl The difference between `cost` and `priority` is that the latter is supposed to get picked irrespective of NEVRA games. It's the closest thing to vendor preferences we have.

However, it's quite likely that the DNF code is _wrong_ for how it is handling the setting at the moment. It is definitely a bug if that's the case.

jkonecny12:
Hello everyone, we think this issue could be solved by this proposal: https://hackmd.io/@prJIqOTeSxC-W0RNGKIGbQ/ByzIz3kIH/edit

Please look and comment there if you are interested.

dwrobel:
@bcl 
> No, the documentation isn't that clear. And from looking at the code in dnf I'm less convinced that this is needed.

Copr has [recently](https://docs.pagure.org/copr.copr/release-notes/2023-08-16.html#possibility-to-specify-a-priority-for-a-repository) added support for specifying the `priority` for repository. Here is the explanation they gave:

>If some package exists in multiple repositories, DNF installs the one with the latest greatest NEVRA, unless repo priority is different. Then, no matter the NEVRA, the package from the most prioritized repo is installed by DNF.

which aligns with my understanding presented [here](https://github.com/pykickstart/pykickstart/pull/275#issuecomment-527537001).

As I also didn't see any outcome from the proposal mentioned [here](https://github.com/pykickstart/pykickstart/pull/275#issuecomment-532213962), unless I missed something.
Would you then reconsider adding the `--priority` option?

I'm using priority extensively [here](https://pagure.io/fork/dwrobel/fedora-kickstarts/blob/f38/f/fedora-disk-base.ks#_100) and I'm not sure how trustful are Copr statistics but quite a few people using it _(~13k is sum of downloads for F36-F38, see [here](https://copr.fedorainfracloud.org/coprs/dwrobel/kernel-rpi/))_. It would be nice not to maintain `pykickstart` fork.

bcl:
If anaconda wants to implement it I will reconsider adding it, but unless they sign off on it I don't want to make things more complicated. I did just try taking a quick look at dnf5's use of it and couldn't untangle how it is used, so maybe it has more of an effect there and would make more sense after dnf5 is the default.

Conan-Kudo:
Priority definitely works in DNF v4, since it's used in Fedora Asahi Remix and CentOS Hyperscale. So, I don't think the option does nothing now.

Conan-Kudo:
From a solver testcase from CentOS Hyperscale, it shows that the priority propagates to libsolv, which uses it to sort and filter packages for installation candidates:

Excerpt from `testcase.t`
```
repo @System -99.-1000 testtags @System.repo.gz
repo centos-hyperscale -50.-1000 testtags centos-hyperscale.repo.gz
repo centos-hyperscale-experimental -50.-1000 testtags centos-hyperscale-experimental.repo.gz
repo centos-hyperscale-spin -50.-1000 testtags centos-hyperscale-spin.repo.gz
repo baseos -99.-1000 testtags baseos.repo.gz
repo appstream -99.-1000 testtags appstream.repo.gz
repo crb -99.-1000 testtags crb.repo.gz
repo extras-common -99.-1000 testtags extras-common.repo.gz
repo epel -99.-1000 testtags epel.repo.gz
repo epel-next -99.-1000 testtags epel-next.repo.gz
system x86_64 rpm @System
```

dwrobel:
>If anaconda wants to implement it I will reconsider adding it, but unless they sign off on it I don't want to make things more complicated.

I'm not sure why this needs to depend on `anaconda`?

To make an image _(using e.g. appliance-creator)_ the missing support for `priority` is in only in:
 - `pykickstart`
 - `livce-tools` - the PR for supporting `priority` is prepared [here](https://github.com/livecd-tools/livecd-tools/pull/134) and marked as blocked by this PR.

dwrobel:
> I don't want to make things more complicated.

Based on my experience, thinks are now more complicated then they should be.
In order to make an image with one package with lower version _(LTS kernel)_ from another repository, one to have fork two aforementioned packages.

bcl:
> So, I don't think the option does nothing now.

Right, but it doesn't do anything more than cost. If you look at the dnf4 code it makes a tuple of them and sorts the repos using that. So as far as I can tell cost can be used to accomplish the same effect.



> I'm not sure why this needs to depend on `anaconda`?

Because anaconda is the primary user of pykickstart, and having options available that don't work in Anaconda will lead to confused users.

Also, I'm not sure a fork is needed to change things. You can extend how pykickstart acts in your application by overriding the commands with your own. eg. https://github.com/rhinstaller/anaconda/blob/master/pyanaconda/kickstart.py#L190

Conan-Kudo:
> Right, but it doesn't do anything more than cost. If you look at the dnf4 code it makes a tuple of them and sorts the repos using that. So as far as I can tell cost can be used to accomplish the same effect.

`cost` does NOT do the same thing. `cost` doesn't exist at the libsolv level, meaning that it has no impact on the filter of available packages.

bcl:
> `cost` does NOT do the same thing. `cost` doesn't exist at the libsolv level, meaning that it has no impact on the filter of available packages.

Sorry about that, I wasn't looking deep enough previously. So I'm more supportive of this, but Anaconda still needs to agree to implement it.

Conan-Kudo:
If it's supported in pykickstart, I'll look into writing the code to wire in the feature to Anaconda.

Conan-Kudo:
@jkonecny12 can we please have agreement to support this? This feature is really going to be needed now that we have Remixes that depend on `priority` to work properly.

jkonecny12:
Sorry for late reply. Unfortunately, we don't have capacity to work on this. Currently the team is highly overloaded. 😔

Conan-Kudo:
What if I did the work, would someone be able to review it?

jkonecny12:
I would say it depends on complexity and test coverage.

PR #278, opened 2019-09-16
Skip FailsToOpenOutputFile_TestCase if running as root

jrodcla:
The expected failure will only happen for non-root users.


jkonecny12:
Jenkins, it's ok to test.

jkonecny12:
Sorry my bad :(

PR #280, opened 2019-09-18
Add note for the repo and url ssl options

jkonecny12:
Tell users that the ssl parameters should not be used and there is preparation for a replacement.

This is a replacement PR for https://github.com/pykickstart/pykickstart/pull/277 .


jkonecny12:
@bcl could you please look on this PR?

PR #282, opened 2019-10-18
Add packages ignore broken parameter

jkonecny12:
This option will disable DNF strict option which will result in ignoring conflicting packages. This option should be used only if there is really a good reason. See the documentation change.

*Related: rhbz#1642013*


pykickstart/sections.py
```diff
@@ -683,6 +684,21 @@ def _getParser(self):
                         Recommends and Supplements flags. By default weak
                         dependencies will be installed.""")
 
+        if self.version < F32:
+            return op
+
+        op.add_argument("--ignorebroken", action="store_true", default=False, version=F32,
+                        help="""
+                        Ignore any packages, groups or modules with conflicting files.
+                        This issue will practically disable DNF strict option.
```
bcl: This will disable the DNF 'strict' config option.

pykickstart/sections.py
```diff
@@ -683,6 +684,21 @@ def _getParser(self):
...
+                        The default behavior will abort the installation with error
```
bcl: The default behavior is to abort the installation ...

pykickstart/sections.py
```diff
@@ -683,6 +684,21 @@ def _getParser(self):
...
+                        message describing the conflicting files.
+
+                        **WARNING: Usage of this parameter is DISCOURAGED! The DNF
+                        strict option will NOT log any information about what packages
+                        were skipped. The resulting system may be in the undefined behavior!**
```
bcl: Replace last sentence with "Using this option may result in an unusable system."

jkonecny12:
**UPDATED**
Implemented notes above. Thanks a lot for your suggestions!
I've changed the last line to ``unstable or unusable`` to make potential stability issue more visible.

----------------------
To answer your question above. I'm discussing with everyone to find out if we want to have this option in RHEL-9. If yes, than the documentation will definitely mark installation with this option as unsupported. 
For Fedora use-case the biggest warning will be here the documentation and we will make a big mark in our logs in the Anaconda.

jkonecny12:
Rebased

PR #285, opened 2019-11-11
Drop python2 compatibility

jkonecny12:
Right now on Rawhide (F32) pykickstart can't be build because
python2-ordered-set and python2-requests do not exists anymore.

The best approach seems to me to drop python2 compatibility and support
pykickstart only as python3 from now on.

---------------------

I don't insist on this solution but it would be great to make the spec file in the upstream working. What do you think @bcl ?

### DIFF
--- a/pykickstart.spec.in
+++ b/pykickstart.spec.in
@@ -1,9 +1,3 @@
-%if 0%{?rhel} > 7
-%bcond_with python2
-%else
-%bcond_without python2
-%endif
-
 # Disable tests by default because they fail to run inside mock builds
 # at the moment, but can run locally.  To build and run tests, do:
 #     rpmbuild -ba --with runtests pykickstart.spec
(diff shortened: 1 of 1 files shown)


bcl:
@jkonecny12 Where and how is this failing? I just did a f32 build and it was successful:

https://koji.fedoraproject.org/koji/buildinfo?buildID=1409125

The python2 build is conditional and we don't build it on anything after rhel7.

jkonecny12:
Maybe I'm mistaken but based on Fedora rpms you are not using your upstream spec file.

https://src.fedoraproject.org/rpms/pykickstart/blob/master/f/pykickstart.spec

bcl:
Ah, ok i think I see now. The spec was actually dropped completely because it really should be maintained in dist-git, not upstream. But was revived for this 'packit' service (which I am not a fan of -- their artifacts have no place in upstream). So for now I'm ok with this change, but to be clear the *right* pykickstart.spec to use is the one from dist-git and this one may go away again at some point in the future.

PR #287, opened 2019-11-11
Fix check for python zanata client in Makefile

jkonecny12:
Makefile should not check specific package because the package name can change and it also changed.

Zanata client is named now `python3-zanata-client` and not `zanata-python-client` as was in Makefile.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,7 @@ NOSEARGS    = -s -v -I __init__.py -I baseclass.py --processes=-1 --process-time
 
 PYTHON_VERSION = $(shell ${PYTHON} -c "print(__import__('sys').version_info[0])")
 
+ZANATA=zanata
 ZANATA_PULL_ARGS = --transdir ./po/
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -18,7 +19,7 @@ all:
 	$(MAKE) -C po
 
 po-pull:
-	rpm -q --whatprovides zanata-python-client &>/dev/null || ( echo "need to run: dnf install zanata-python-client"; exit 1 )
+	which $(ZANATA_CLIENT_BIN) &>/dev/null || ( echo "need to install zanata python client package"; exit 1 )
 	zanata pull $(ZANATA_PULL_ARGS)
```
bcl: If we're going to use a variable, it should be used everywhere. I'd also name it $ZANATA so it isn't so long.

jkonecny12: So true. Fixed.

jkonecny12:
**UPDATED**

PR #290, opened 2019-11-21
Add the rhsm command

M4rtinK:
Add a new command called "rhsm", that makes it possible to specify
Red Hat subscription related data, used to subscribe a system during
a kickstart based installation and to enable subscription related services.

Options at a glance:

--organization & --activation-key (can be passed more than once)

Organization id and at least once activation key are required.

--connect-to-insights

Connects the target system to Red Hat Insights.

--proxy

Configures RHSM to use a HTTP proxy, syntax is the same as for
other --proxy options for other commands, such as the repo command.

--server-hostname

Sets custom Candlepin instance URL.

--rhsm-baseurl

Set custom content base URL.

Related: rhbz#1748281

### DIFF
--- a/pykickstart/commands/rhsm.py
+++ b/pykickstart/commands/rhsm.py
@@ -1,0 +1,112 @@
+#
+# Copyright 2019 Red Hat, Inc.
+#
+# This copyrighted material is made available to anyone wishing to use, modify,
+# copy, or redistribute it subject to the terms and conditions of the GNU
+# General Public License v.2.  This program is distributed in the hope that it
+# will be useful, but WITHOUT ANY WARRANTY expressed or implied, including the
+# implied warranties of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+# See the GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along with
+# this program; if not, write to the Free Software Foundation, Inc., 51
+# Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.  Any Red Hat
+# trademarks that are incorporated in the source code or documentation are not
+# subject to the GNU General Public License and may only be used or replicated
+# with the express permission of Red Hat, Inc.
+#
+
+from pykickstart.base import KickstartCommand
+from pykickstart.errors import KickstartParseError
+from pykickstart.options import KSOptionParser
+from pykickstart.version import RHEL8
+
+from pykickstart.i18n import _
+
+class RHEL8_RHSM(KickstartCommand):
+    removedKeywords = KickstartCommand.removedKeywords
+    removedAttrs = KickstartCommand.removedAttrs
+
+    def __init__(self, writePriority=0, *args, **kwargs):
+        KickstartCommand.__init__(self, writePriority, *args, **kwargs)
+        self.op = self._getParser()
+
+        self.organization = kwargs.get("organization", None)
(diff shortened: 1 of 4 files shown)


M4rtinK:
Anaconda PR: https://github.com/rhinstaller/anaconda/pull/2218

tests/commands/rhsm.py
```diff
@@ -0,0 +1,77 @@
+#
+# Martin Kolman <mkolman@redhat.com>
```
bcl: No need for names anymore, that's what git blame is for :)

pykickstart/commands/rhsm.py
```diff
@@ -0,0 +1,118 @@
+#
+# Martin Kolman <mkolman@redhat.com>
```
bcl: No need for names.

pykickstart/commands/rhsm.py
```diff
@@ -0,0 +1,118 @@
...
+                        help="Red Hat subscription service server URL.")
+        op.add_argument("--rhsm-baseurl", metavar="<content_base_url>", version=RHEL8, required=False,
+                        help="Content base URL.")
+        return op
+
+    def parse(self, args):
+        (ns, extra) = self.op.parse_known_args(args=args, lineno=self.lineno)
+
+        if len(extra) > 0:
+            msg = _("Unsupported arguments specified for the rhsm command: %s" % extra)
```
bcl: Move the % extra outside the translation _() brackets.
pylint generates a warning about this, which isn't obvious unless you look at the test run logs.

pykickstart/commands/rhsm.py
```diff
@@ -0,0 +1,118 @@
...
+                        ``--proxy=[protocol://][username[:password]@]host[:port]``
+                        """)
+        op.add_argument("--server-hostname", metavar="<subscription_service_server_url>", version=RHEL8, required=False,
+                        help="Red Hat subscription service server URL.")
+        op.add_argument("--rhsm-baseurl", metavar="<content_base_url>", version=RHEL8, required=False,
+                        help="Content base URL.")
+        return op
+
+    def parse(self, args):
+        (ns, extra) = self.op.parse_known_args(args=args, lineno=self.lineno)
```
poncovka: Please, use `parse_args` if you don't expect extra arguments.

tests/commands/rhsm.py
M4rtinK: OK, removed. :)

pykickstart/commands/rhsm.py
M4rtinK: Removed.

pykickstart/commands/rhsm.py
M4rtinK: Good catch! :) Fixed.

pykickstart/commands/rhsm.py
M4rtinK: Good point, changed. I guess we can drop the extra handling then & `parse_args()` already has some error handling code for that ?

M4rtinK:
@bcl Indeed, this definitely needs to go upstream to be available for future RHEL releases. I guess I'll open an upstream PR once this one is merged, so that we can avoid keeping the two PRs in sync ?

pykickstart/commands/rhsm.py
poncovka: Yes.

PR #294, opened 2019-12-06
Add lowercase aliases for camelCased %packages options

VladimirSlavik:
The majority of Kickstart options use lowercase hyphenated naming scheme. This adds the `--exclude-weakdeps` alias for `--excludeWeakdeps` and `--inst-langs` for `--instLangs`, which moves Kickstart a bit closer to consistency.

### DIFF
--- a/pykickstart/commands/timezone.py
+++ b/pykickstart/commands/timezone.py
@@ -17,9 +17,10 @@
 # subject to the GNU General Public License and may only be used or replicated
 # with the express permission of Red Hat, Inc.
 #
+import warnings
 from pykickstart.version import FC3, FC6, F18
 from pykickstart.base import KickstartCommand
-from pykickstart.errors import KickstartParseError
+from pykickstart.errors import KickstartParseError, KickstartDeprecationWarning
 from pykickstart.options import KSOptionParser, commaSplit
 
 from pykickstart.i18n import _
@@ -383,5 +384,39 @@ class F25_Timezone(F23_Timezone):
         if self.ntpservers and self.nontp:
             msg = _("Options --nontp and --ntpservers are mutually exclusive")
             raise KickstartParseError(msg, lineno=self.lineno)
+
+        return self
+
+class F32_Timezone(F25_Timezone):
+
+    def __init__(self, writePriority=0, *args, **kwargs):
+        F25_Timezone.__init__(self, writePriority, *args, **kwargs)
+        self.op = self._getParser()
+
+    def _getArgsAsStr(self):
+        retval = ""
+
+        if self.timezone:
+            retval += " " + self.timezone
+
+        if self.isUtc:
+            retval += " --utc"
+
+        if self.nontp:
+            retval += " --nontp"
+
+        if self.ntpservers:
+            retval += " --ntpservers=" + ",".join(self.ntpservers)
+
+        return retval
+
+    def parse(self, args):
+        F25_Timezone.parse(self, args)
+
+        if "--isUtc" in args:
+            warnings.warn(_("The option --isUtc will be deprecated in future releases. Please "
+                            "modify your kickstart file to replace this option with its preferred "
+                            "alias --utc."),
+                          KickstartDeprecationWarning)
 
         return self
--- a/pykickstart/commands/timezone.py
+++ b/pykickstart/commands/timezone.py
@@ -290,6 +290,32 @@ class F25_Timezone(F23_Timezone):
         F23_Timezone.__init__(self, writePriority, *args, **kwargs)
         self.op = self._getParser()
 
+    def __str__(self):
+        retval = KickstartCommand.__str__(self)
+        args = self._getArgsAsStr()
+        if args:
+            retval += "# System timezone\n"
+            retval += "timezone" + args + "\n"
+
+        return retval
+
+    def _getArgsAsStr(self):
+        retval = ""
+
+        if self.timezone:
+            retval += " " + self.timezone
+
+        if self.isUtc:
+            retval += " --isUtc"
+
+        if self.nontp:
+            retval += " --nontp"
+
+        if self.ntpservers:
+            retval += " --ntpservers=" + ",".join(self.ntpservers)
+
+        return retval
+
     def _getParser(self):
         op = KSOptionParser(prog="timezone", description="""
                             This required command sets the system time zone to
--- a/pykickstart/sections.py
+++ b/pykickstart/sections.py
@@ -29,11 +29,12 @@ treated just the same as a predefined one by the kickstart parser.  All that
 is necessary is to create a new subclass of Section and call
 parser.registerSection with an instance of your new class.
 """
+import warnings
 from pykickstart.constants import KS_SCRIPT_PRE, KS_SCRIPT_POST, KS_SCRIPT_TRACEBACK, \
                                   KS_SCRIPT_PREINSTALL, KS_SCRIPT_ONERROR, \
                                   KS_MISSING_IGNORE, KS_MISSING_PROMPT, \
                                   KS_BROKEN_IGNORE, KS_BROKEN_REPORT
-from pykickstart.errors import KickstartParseError
+from pykickstart.errors import KickstartParseError, KickstartDeprecationWarning
 from pykickstart.options import KSOptionParser
 from pykickstart.version import FC4, F7, F9, F18, F21, F22, F24, F32, RHEL6, RHEL7
 from pykickstart.i18n import _
@@ -766,3 +767,12 @@ class PackageSection(Section):
             self.handler.packages.handleBroken = KS_BROKEN_IGNORE
         else:
             self.handler.packages.handleBroken = KS_BROKEN_REPORT
+
+        for option, new_option in \
+                {"--instLangs": "--inst-langs", "--excludeWeakdeps": "--exclude-weakdeps"}.items():
+            if option in args:
+                warnings.warn(_("The %(option)s option on line %(lineno)s will be deprecated in"
+                                "future releases. Please modify your kickstart file to replace"
+                                "this option with its preferred alias %(new_option)s.")
+                              % {"option": option, "lineno": lineno, "new_option": new_option},
+                              KickstartDeprecationWarning)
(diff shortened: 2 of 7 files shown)


pykickstart/sections.py
```diff
@@ -676,7 +676,7 @@ def _getParser(self):
         if self.version < F24:
             return op
 
-        op.add_argument("--excludeWeakdeps", dest="excludeWeakdeps",
+        op.add_argument("--excludeWeakdeps", "--exclude-weakdeps", dest="excludeWeakdeps",
```
poncovka: Could you add tests for these two options, please?

bcl:
I don't think this is really necessary. It's unfortunate that it has been inconsistent, but I think adding aliases it just going to lead to confusion down the line.
What might be neat is to add a pylint test that check for camelCase and complains.

pykickstart/sections.py
bcl: I agree. And it looks like the existing tests may need a little work, I don't see it testing --excludeWeakdeps on the input side, only that it is added by setting pkgs.excludeWeakdeps

If we are going to make this change we should also change the output so that it uses the new form. And while I'm thinking of things to do :) maybe it would be best to add the alias, and deprecate the old version so we can remove it later? Ideally this would all be done in the F32 part of PackageSection instead of where it was originally defined. Looking at the existing code I'm not sure we've done something like that before.

bcl:
Thanks for the new test! That's exactly what I was thinking.

VladimirSlavik:
@bcl @poncovka So, this should be roughly ready for review - the overall shape is now in place. Let me recap what this should change, high-level:
* Add non-camelCase aliases for any camelCased options. (Turns out there are only 2 in %packages and 1 in timezone.)
* Use the new aliases in string casting from command classes.
* Warn when these camelCase options are found.
* Make all of this happen starting with F32 version.
* Add tests for this.
* Add another test for available options names (to warn devs while implementing new options).

Given I'm rather new at pykickstart and indeed even modern Python, there's a bunch of open questions I can see right now:
* Should I split the timezone changes to another PR?
* Use warning for the camelCase occurrences, or something else?
* What about the checks for version - is it ok as done now with isinstance?

I'm sure there is more...

tests/args_names.py
```diff
@@ -0,0 +1,87 @@
+import unittest
+import shlex
+from re import compile
```
bcl: This conflicts with the builtin compile, just import re and use re.compile

pykickstart/commands/timezone.py
```diff
@@ -116,18 +117,24 @@ def __init__(self, writePriority=0, *args, **kwargs):
 
     def __str__(self):
```
bcl: Instead of populating this with isinstance checks let's just make a new one for the F32 class. Keeps the logic cleaner and makes it easier to read.

pykickstart/commands/timezone.py
```diff
@@ -116,18 +117,24 @@ def __init__(self, writePriority=0, *args, **kwargs):
...
-        if self.timezone:
+        args = self._getArgsAsStr()
+        if self.timezone or (isinstance(self, F25_Timezone) and args):
             retval += "# System timezone\n"
-            retval += "timezone " + self._getArgsAsStr() + "\n"
+            retval += "timezone" + args + "\n"
 
         return retval
 
     def _getArgsAsStr(self):
```
bcl: Same thing here, just make a new one for the new class.

tests/args_names.py
```diff
@@ -0,0 +1,87 @@
...
+
+    def runTest(self):
+
+        self._check_commands()
+        self._check_sections()
+        self._report()
+
+    def _check_sections(self):
+
+        sections = {
```
bcl: The section openers should be populated from the class .sectionOpen instead of hard-coded here. Sure, they're not likely to change, but that will mean less things to change when adding new ones to check.

tests/args_names.py
```diff
@@ -0,0 +1,87 @@
...
+
+            found_any_good = False
+            for option in action.option_strings:
+                if cmd_class and option.lstrip("-") in cmd_class.removedAttrs:
+                    # caution removedAttrs does not include leading dashes
+                    continue
+                is_good = self._reg_name_checker.fullmatch(option) is not None
+                found_any_good = found_any_good or is_good
+                if not is_good:
+                    print("Found option with uppercase letters: {} for command {}.".format(
```
bcl: Let's try not to use .format() unless it really adds something that '%' doesn't do. In general pykickstart has only used %s and %(dictname)s in the string formatting so let's stick with that.
Also, for long lines, I like to err on the side of readability vs. wrapping at someone else's edit width so I don't think this needs to be on 2 lines.

tests/args_names.py
```diff
@@ -0,0 +1,87 @@
...
+                        option, cmd_name))
+
+            if not found_any_good:
+                self._bad_options.append(tuple((cmd_name, action.option_strings)))
+
+    def _report(self):
+
```
bcl: extra empty line can be removed.

tests/args_names.py
```diff
@@ -0,0 +1,87 @@
...
+        if self._bad_options:
+            print("The following kickstart option sets do not include a lowercase only variant:")
+            for option_set in self._bad_options:
+                print("{}: {}".format(option_set[0], ", ".join(option_set[1])))
```
bcl: Replace format() with %

tests/commands/timezone.py
```diff
@@ -154,5 +155,47 @@ def runTest(self):
         self.assert_parse_error("timezone Europe/Sofia --nontp --ntpservers=0.fedora.pool.ntp.org,1.fedora.pool.ntp.org")
 
 
+class F32_TestCase(F25_TestCase):
+    command = "timezone"
+
+    def runTest(self):
+
```
bcl: extra empty line can go.

tests/commands/timezone.py
```diff
@@ -154,5 +155,47 @@ def runTest(self):
...
+        # Failures
+        # unknown argument
+        self.assert_parse_error("timezone --blah")
+        # more than two timezone specs
+        self.assert_parse_error("timezone foo bar", KickstartParseError,
+                                'One or zero arguments are expected for the timezone command')
+        self.assert_parse_error("timezone --utc foo bar", exception=KickstartParseError)
+        # no options
+        self.assert_parse_error("timezone", KickstartParseError,
+                    'At least one option and/or an argument are expected for the timezone command')
```
bcl: When wrapping lines like this, start the next line under the function ( like you did for line 168.

tests/commands/timezone.py
```diff
@@ -154,5 +155,47 @@ def runTest(self):
...
+        # contradictory options
+        self.assert_parse_error(
```
bcl: It's already really long, may as well not wrap it.

tests/packages.py
```diff
@@ -552,5 +552,23 @@ def runTest(self):
             self.assertEqual(len(w), 0)
             self.assertEqual(str(self.handler.packages), "\n%packages --retries=10\nbash\n\n%end\n")
 
+class Packages_Warn_CamelCase_TestCase(ParserTest):
+    def __init__(self, *args, **kwargs):
+        ParserTest.__init__(self, *args, **kwargs)
+        self.version = F32
+        self.ks = """
```
bcl: May as well make this all one line with embedded \n like you did in the assert check.

tests/commands/timezone.py
```diff
@@ -154,5 +155,47 @@ def runTest(self):
...
+        # Successes
+        # normal contents
+        self.assert_parse("timezone Europe/Prague --ntpservers=ntp.cesnet.cz",
+                          "timezone Europe/Prague --ntpservers=ntp.cesnet.cz\n")
+        # no timezone spec
+        self.assert_parse("timezone --ntpservers=ntp.cesnet.cz",
+                          "timezone --ntpservers=ntp.cesnet.cz\n")
+
+        # New in F32: any variant of UTC should be returned as --utc (again - was in FC3)
+        with warnings.catch_warnings(record=True) as w:
```
poncovka: You should be able to use `self.assertWarns` to check the warnings. See: https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertWarns

tests/args_names.py
```diff
@@ -0,0 +1,87 @@
...
+
+            assert False
```
poncovka: You can use `self.fail` to signal a test failure with a message. Btw. the test can fail for the first invalid option. You don't have to collect them all. It could simplify the code a little. See: https://docs.python.org/3/library/unittest.html#unittest.TestCase.fail

VladimirSlavik: I think one of our previous conversations was around providing a full list? Maybe I remembered it wrong.

VladimirSlavik:
Still todo: Clean up the fix for F25_Timezone (or drop?) && decide if the option name test should die early or keep going to catch them all.

bcl:
I think the tests should catch them all, it's much nicer to see them all at ones instead of having to iterate a bunch of times when fixing things.

Looks pretty good, thanks for all the work!

tests/args_names.py
poncovka: It was just a suggestion. It can stay as it is :-)

VladimirSlavik:
Hi @bcl, I fixed the last bit noted above. But I'm at loss at the test failures: I can't reproduce these locally, and the tracebacks aren't really failing tests as far as I can tell. As this popped up only now and worked before, I guess it's some breakage of the testing machinery that happened over the holidays...?

bcl:
Take a look at the coverage report - https://coveralls.io/builds/27940805/source?filename=pykickstart%2Fcommands%2Ftimezone.py#L298

looks like we need to add a few more lines of testing to cover the new code.

VladimirSlavik:
Changed - locally I get now 100% for timezone.py. Hopefully that does the job. Sorry to keep piling up problems :(

bcl:
Thanks! I just want to make sure we don't let things slip through. It looks like the coverage problem is this one, which is being worked on - https://github.com/nedbat/coveragepy/issues/916

VladimirSlavik:
Hi Brian @bcl , do you think you could please make a release with this? Our tests use daily builds so we had to update them once this was merged, but there's no release so we don't have a version to require.

bcl:
Yeah, as soon as we get the webalate stuff sorted out I'll do a new release.

PR #297, opened 2019-12-17
cherry-pick the RHEL only rhsm command to master

bcl:


bcl:
Includes the commits from #296 so that tests pass cleanly.

PR #298, opened 2019-12-18
Add the rhsm command

M4rtinK:
Add a new command called "rhsm", that makes it possible to specify
Red Hat subscription related data, used to subscribe a system during
a kickstart based installation and to enable subscription related services.

Options at a glance:

--organization & --activation-key (can be passed more than once)

Organization id and at least once activation key are required.

--connect-to-insights

Connects the target system to Red Hat Insights.

--proxy

Configures RHSM to use a HTTP proxy, syntax is the same as for
other --proxy options for other commands, such as the repo command.

--server-hostname

Sets custom Candlepin instance URL.

--rhsm-baseurl

Set custom content base URL.

Resolves: rhbz#1776378
Related: rhbz#1748281

### DIFF
--- a/pykickstart/commands/__init__.py
+++ b/pykickstart/commands/__init__.py
@@ -69,6 +69,7 @@ from pykickstart.commands import (
     reboot,
     repo,
     rescue,
+    rhsm,
     rootpw,
     selinux,
     services,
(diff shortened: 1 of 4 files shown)


bcl:
Is this different from #297?

M4rtinK:
Oops, looks like we both decided to cherry pick the RHSM command to master at roughly the same time & me not noticing there is already a PR for this open. :) So I guess this one could be closed then & we can go with #297 ? :)

bcl:
> go with #297 ? :)

yes, since it has the working tests patch as well :)

PR #309, opened 2020-03-13
Remove url command doc that it's replaced

jkonecny12:
URL command have still it's meaning. Not sure why this part of the documentation was added but `mirrorlist` does not replacing url.

Mirrorlist should be used for mirrorlist server, however, url should be used to point directly to the server repository.

### DIFF
--- a/pykickstart/commands/url.py
+++ b/pykickstart/commands/url.py
@@ -175,10 +175,10 @@ class F18_Url(F14_Url):
     def _getParser(self):
         op = F14_Url._getParser(self)
         op.add_argument("--url", version=F18, help="""
-                        This parameter is no longer required because you could
-                        use ``--mirrorlist`` instead.""")
+                        The URL to install from. Variable substitution is done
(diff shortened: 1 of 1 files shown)


jkonecny12:
**UPDATED:**
Great that you have these tests running. I missed that the original command `--url` changed to optional from required.

jkonecny12:
Unfortunately this doesn't look nice in the documentation :(.

https://pykickstart.readthedocs.io/en/latest/kickstart-docs.html#url

VladimirSlavik:
The contents of "help" for overridden params in subclasses are displayed after the "changed in version X" and mostly contain text about the changes. So I'd argue the original text was the more correct one after all.

bcl:
> So I'd argue the original text was the more correct one after all.

No, it was misleading. I think it would be clearer if F18 said "--url or --mirrorlist must be specified." and in F27 say "--url, --mirrorlist, or --metalink much be specified."

VladimirSlavik:
Ah, that makes far more sense.

PR #310, opened 2020-03-18
Make --url doctext more understandable

jkonecny12:
Documentation for --url duplicating between version. Let's take another approach and tell there what changed instead of what it is doing on version update.

Update to fix problems of https://github.com/pykickstart/pykickstart/pull/309.

### DIFF
--- a/pykickstart/commands/url.py
+++ b/pykickstart/commands/url.py
@@ -175,8 +175,7 @@ class F18_Url(F14_Url):
     def _getParser(self):
         op = F14_Url._getParser(self)
         op.add_argument("--url", version=F18, help="""
-                        The URL to install from. Variable substitution is done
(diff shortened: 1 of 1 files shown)


pykickstart/commands/url.py
```diff
@@ -237,6 +236,9 @@ def __str__(self):
 
     def _getParser(self):
         op = F18_Url._getParser(self)
+        op.add_argument("--url", version=F18, help="""
```
poncovka: I guess that there should be the version F27.

jkonecny12: oh my, sorry ...
Fixed.

jkonecny12:
Updated to fix bug pointed by @poncovka

PR #312, opened 2020-03-25
Deprecate repo ignoregroups

jkonecny12:
The ``repo --ignoregroups`` parameter is not used by Anaconda from yum -> dnf transition. I don't think we want to add it back but instead we can remove this from our API.

P.S.: Not sure if the deprecation is done correctly. Could you please verify that before merge @bcl.

### DIFF
--- a/pykickstart/commands/repo.py
+++ b/pykickstart/commands/repo.py
@@ -19,7 +19,7 @@
 #
 from textwrap import dedent
 from pykickstart.version import versionToLongString
-from pykickstart.version import FC6, F8, F11, F13, F14, F15, F21, F27, F30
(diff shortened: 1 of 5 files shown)


bcl:
I've added some fixes as a fixup commit, squash it with the previous one if you like them.

jkonecny12:
Thanks a lot for your fixes.

jkonecny12:
Could this be merged or do we waiting for something?

bcl:
I was waiting for you to squash my fixup commit with yours :)

jkonecny12:
Squashed! Sorry it took me so long.

PR #315, opened 2020-05-12
Update travis to python3.8 and fix pylint 2.5.0 warning about _

bcl:


jkonecny12:
It's reported here https://github.com/PyCQA/pylint/issues/3604.

bcl:
I changed i18n.py which has the potential to screw up translations. I think it is correct, but if someone could double check that would be nice.

Running ksvalidator with `LANG=zh_CN.UTF-8 ksvalidator --help` does return what appears to be translated help strings so I think it is working ok.

bcl:
Merged to master

PR #316, opened 2020-05-13
move dependencies into setup.py and use setuptools

### LINKED ISSUE #268: Pypi Package: Missing install dependencies/requires
Hey,

thank you for maintaining this awesome tool. As per our discussion per e-mail, I will open this issue here.

Installing pykickstart from pypi is missing some dependencies and requirements needed for a proper installation.

I was able to identify:
- six
- requests
- ordered_set

But there may be more.

Having a look at the setup.py, there seems to be some more work to be done.
If it is ok for you, I will start working on the same and provide a PR during the next time.

Regards,
Daniel

CMeza99:
* use `setuptools`
* move dependencies to setup.py so future versions can install from PyPI
* resolves #268


requirements.txt
```diff
@@ -1,16 +1,14 @@
...
 coveralls
-ordered-set
 polib
 pylint
-requests
-six
 Sphinx == 1.7.6
 
 # Custom Red Hat module used with pylint
 https://github.com/rhinstaller/pocketlint/zipball/master
```
jkonecny12, line 11: Pocketlint is also in pypi now. It can be dragged in as the rest of the dependencies.

https://pypi.org/project/pocketlint/

requirements.txt
```diff
@@ -1,16 +1,14 @@
 # Requirements for automatic testing and building
 # Runtime requirements are described in the spec file
 
-pip
-setuptools
 coverage
 coveralls
-ordered-set
 polib
 pylint
```
Conan-Kudo, line 7: You can put the test dependencies as `test_requires` in `setup.py`.

bcl:
Please leave requirements.txt as-is, it is useful for setting up pip in a venv, container, or travis.
I'm fine with moving to setuptools and including deps in setup.py though.

sagarun:
Are you still interested in getting this merged? I am interested in this PR. I can take over it if you like @CMeza99

requirements.txt
CMeza99: The `test` command was deprecated in `setuptools` v41.5.0. Which means, so has `test_requires`. See https://setuptools.readthedocs.io/en/latest/setuptools.html

I suggest using `extra requires` and creating a group called "test".

requirements.txt
```diff
@@ -1,16 +1,2 @@
-# Requirements for automatic testing and building
```
sagarun, line 1: @CMeza99 I believe @bcl  requested to leave the requirements.txt alone for use in other areas

`Please leave requirements.txt as-is, it is useful for setting up pip in a venv, container, or travis.
I'm fine with moving to setuptools and including deps in setup.py though.`

CMeza99:
@sagarun thanks for the ping. i did not mean to abandon this. :)
@bcl the `requirements.txt` now references what is in setup.py. This way deps are not declared in three places causing inconsistencies (i.e. `nose` is still declared in `pykickstart.spec.in`). Existing scripts and workflows should not be impacted.

requirements.txt
CMeza99: I was still commenting when you wrote this. I responded to the concern below. :)

setup.py
```diff
@@ -1,28 +1,13 @@
...
-
-setup(cmdclass={"install_scripts": install_scripts},
-      name='pykickstart', version='3.27',
+setup(name='pykickstart', version='3.27',
       description='Python module for manipulating kickstart files',
       author='Chris Lumens', author_email='clumens@redhat.com',
       url='http://fedoraproject.org/wiki/pykickstart',
+      install_requires=['six', 'ordered-set', 'requests'],
+      extras_require={
+          "docs": ['Sphinx'],
```
sagarun, line 9: sphinx was pinned to ``Sphinx == 1.7.6`` in requirements.txt. It is no longer in setup.py JFYI

CMeza99: My thinking was that was an old artifact. Here it looks like the version of Sphinx is dependent on what the OS installs, `https://github.com/pykickstart/pykickstart/blob/master/pykickstart.spec.in#L25`. 

FYI: Sphinx is currently at 3.1. 1.7.6 was released Jul 17, 2018. The last `Sphinx < 2` is 1.8.5 released Mar 10, 2019.

requirements.txt
```diff
@@ -1,16 +1,2 @@
...
-polib
-pylint
-requests
-six
-Sphinx == 1.7.6
-
-# Custom Red Hat module used with pylint
-https://github.com/rhinstaller/pocketlint/zipball/master
+# Installs all extras (for testing and documentation) in pykickstart.
+.[docs,test]
```
bcl, line 2: Ok, that's pretty darn neat.

setup.py
```diff
@@ -1,28 +1,13 @@
...
-                os.unlink(new)
-            move_file(old, new)
-
-setup(cmdclass={"install_scripts": install_scripts},
-      name='pykickstart', version='3.27',
+setup(name='pykickstart', version='3.27',
       description='Python module for manipulating kickstart files',
       author='Chris Lumens', author_email='clumens@redhat.com',
       url='http://fedoraproject.org/wiki/pykickstart',
+      install_requires=['six', 'ordered-set', 'requests'],
```
bcl, line 7: Remove `ordered-set` and rebase it on master after I push #330 and I think this one will be good.

CMeza99:
Is there anything else I need to do to get this merged?

PR #317, opened 2020-05-13
add testing for py38

CMeza99:
* add `py38` (allowed to fail)
* update travis-ci syntax
* update dist to bionic


bcl:
Thanks, but I'm already working on 3.8 in #315

Issue #307, opened 2020-02-25
3.23: build documanetation issues


kloczek:
Looks like there are minor some issue with building documentation.

Other thing is that docs target is fetching some fedora documentation files.
It makes a difficult when documentation is build in network without connectivity to public network.

```
[tkloczko@barrel pykickstart-r3.23]$ make docs
fatal: not a git repository (or any parent up to mount point /home/tkloczko)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
make -C docs html text
make[1]: Entering directory '/home/tkloczko/rpmbuild/BUILD/pykickstart-r3.23/docs'
sphinx-apidoc -o . ../pykickstart ; \

Creating file ./pykickstart.rst.
Creating file ./pykickstart.commands.rst.
Creating file ./pykickstart.handlers.rst.
Creating file ./modules.rst.
./gen_commands_docs > commands.rst
./gen_commands_docs RHEL8 > commands_rhel.rst
./gen_sections_docs > sections.rst
STARTED_FROM_MAKEFILE=1 sphinx-build -b html -d _build/doctrees   . _build/html
Running Sphinx v2.2.2
making output directory... done
WARNING: html_static_path entry '_static' does not exist
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 10 source files that are out of date
updating environment: [new config] 10 added, 0 changed, 0 removed
reading sources... [100%] testing
/home/tkloczko/rpmbuild/BUILD/pykickstart-r3.23/docs/commands.rst:719: WARNING: Explicit markup ends without a blank line; unexpected unindent.
/home/tkloczko/rpmbuild/BUILD/pykickstart-r3.23/docs/commands.rst:731: WARNING: Explicit markup ends without a blank line; unexpected unindent.
/home/tkloczko/rpmbuild/BUILD/pykickstart-r3.23/docs/commands.rst:1633: WARNING: Explicit markup ends without a blank line; unexpected unindent.
/home/tkloczko/rpmbuild/BUILD/pykickstart-r3.23/docs/commands.rst:2082: WARNING: Explicit markup ends without a blank line; unexpected unindent.
/home/tkloczko/rpmbuild/BUILD/pykickstart-r3.23/docs/commands.rst:2302: WARNING: Explicit markup ends without a blank line; unexpected unindent.
[...]

PR #319, opened 2020-05-13
docs: Maintain programmers-guide in the repo

bcl:
This was being pulled from the Fedora wiki, but has remained unchanged
since it was first posted. pykickstart docs are now separate from Fedora
so it is easier to maintain this here in the repo.

This also removes the need for network access while building, which is
usually not available.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,7 +33,6 @@ po-push:
 
 docs:
 	$(MAKE) -C docs html text
-	curl -A "programmers-guide" -o docs/programmers-guide "https://fedoraproject.org/w/index.php?title=PykickstartIntro&action=raw"
 
 check:
 ifneq ($(PYTHON_VERSION),3)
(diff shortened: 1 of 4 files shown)


bcl:
> Incidentally, it'd probably be worth double checking how much of this document is still relevant and correct.

Yeah, I figured I'd do that at some point as well -- the wiki copy hasn't been changed in almost exactly forever :)

bcl:
All these together should fix #307

VladimirSlavik:
Maybe you could remove the wiki page, once the doc is published? Realistically, only you (pykickstart developers) can make meaningful changes to this text anyway, regardless of where and how it is published.

bcl:
> Maybe you could remove the wiki page, once the doc is published?

I'll update it to point to the file in the github repo so that people following search results don't get lost :)

PR #322, opened 2020-06-19
Add the timesource command

M4rtinK:
With the effort to add support for NTS (secure NTP) to Fedora
we have found out the current way of specifying NTP servers
via the --ntpservers option of the timezone command is no longer
sufficient. Not only can't it be easily and cleanly extended to also
support NTS servers, we are actually missing some options that
are available in the GUI, such as NTP pool support.

So add a new command called "timesource" that makes it possible to add
multiple NTP servers/pools and to specify if the given server/pool
supports NTS.

This way we can add support for NTS in a robust manner & make it easy
to add support for other time synchronization features as needed.

The timesource command can be used multiple times, one NTP/NTS server/pool
per line.

Options at a glance:

--ntp-server

Address of a NTP server.

--ntp-pool

Address of a NTP pool.

NOTE: It is not possible to specify both --ntp-server and --ntp-pool at the same time,
as the result would be ambiguous.

--nts

Specifies that the given server or pool supports the NTS protocol.

--ntp-disable

Specifies that all ntp based time synchronization should be disabled on the system.
If timesource is specified multiple times and at least one invocation uses the
--ntp-disable option, ntp based time synchronization will be disabled.

**TODO:**
~- mark the `--ntpservers` option of the `timezone` command as deprecated (it will still be parsed by Anaconda for the foreseeable future, but users should be aware it is deprecated)~


pykickstart/commands/ntp.py
```diff
@@ -0,0 +1,137 @@
...
+    removedKeywords = KickstartCommand.removedKeywords
+    removedAttrs = KickstartCommand.removedAttrs
+
+    def __init__(self, writePriority=0, *args, **kwargs):
+        KickstartCommand.__init__(self, writePriority, *args, **kwargs)
+        self.op = self._getParser()
+        self.NTPList = kwargs.get("NTPList", [])
+        self.exclusive_required_options = [("server", "--server"),
+                                           ("pool", "--pool")]
+
```
jkonecny12, line 81: Please remove this empty line.

pykickstart/commands/ntp.py
```diff
@@ -0,0 +1,137 @@
...
+    def _getParser(self):
+        op = KSOptionParser(prog="ntp", description="""
+                            Configures the Network Time Protocol.""",
+                            version=F33)
+        op.add_argument("--server", version=F33, help="""
+                        A single NTP server.
+
+                        ``--server`` and ``--pool`` are mutually exclusive.
+                        """,
+                        metavar="<group1>,<group2>,...,<groupN>")
```
jkonecny12, line 99: I don't think we want to have multiple servers on one --server.

pykickstart/commands/ntp.py
```diff
@@ -0,0 +1,137 @@
...
+        if self.server:
+            return self.server == y.server and self.nts == y.nts
+        elif self.pool:
+            return self.pool == y.pool and self.nts == y.nts
+
+    def __ne__(self, y):
+        return not self == y
+
+    def __str__(self):
+        retval = BaseData.__str__(self)
```
poncovka, line 50: Please, use `super()`.

pykickstart/commands/ntp.py
```diff
@@ -0,0 +1,137 @@
...
+        for ntp_command in self.NTPList:
+            retval += ntp_command.__str__()
+
+        return retval
+
+    def _getParser(self):
+        op = KSOptionParser(prog="ntp", description="""
+                            Configures the Network Time Protocol.""",
+                            version=F33)
+        op.add_argument("--server", version=F33, help="""
```
poncovka, line 94: We should also support the `--disabled` option.

pykickstart/commands/ntp.py
```diff
@@ -0,0 +1,137 @@
...
+            retval += " --server=%s" % self.server
+        if self.pool:
+            retval += " --pool=%s" % self.pool
+        if self.nts:
+            retval += " --nts"
+
+        return retval
+
+
+class F33_NTP(KickstartCommand):
```
poncovka, line 71: Please, create also `F33_Timezone` and issue `KickstartDeprecationWarning` for the `--nontp` and `--ntpservers` options. There options will be deprecated in the future.

pykickstart/commands/ntp.py
VladimirSlavik, line 99: 1) If I read this right, a line like this is valid: `ntp --server=time.example.com,time.windows.com,ntp.cesnet.cz`. IIRC the idea was to use one line per host, which this does... not?

2) I would expect `<server1>, <server2>,...,<serverN>` instead of "group" there...

3) You don't test for this advertised usage ;-)

pykickstart/commands/ntp.py
M4rtinK: Removed.

pykickstart/commands/ntp.py
M4rtinK: Copy paste error fixed.

pykickstart/commands/ntp.py
```diff
@@ -0,0 +1,137 @@
...
+from pykickstart.options import KSOptionParser
+
+from pykickstart.i18n import _
+
+class F33_NTPData(BaseData):
+    removedKeywords = BaseData.removedKeywords
+    removedAttrs = BaseData.removedAttrs
+
+    def __init__(self, *args, **kwargs):
+        BaseData.__init__(self, *args, **kwargs)
```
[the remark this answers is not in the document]
M4rtinK: Changed.

pykickstart/commands/ntp.py
```diff
@@ -0,0 +1,137 @@
...
+        self.server = kwargs.get("server", [])
```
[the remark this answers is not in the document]
M4rtinK: Copy paste error - fixed now.

pykickstart/commands/ntp.py
M4rtinK: Changed.

pykickstart/commands/ntp.py
```diff
@@ -0,0 +1,137 @@
...
+
+        return retval
+
+
+class F33_NTP(KickstartCommand):
+    removedKeywords = KickstartCommand.removedKeywords
+    removedAttrs = KickstartCommand.removedAttrs
+
+    def __init__(self, writePriority=0, *args, **kwargs):
+        KickstartCommand.__init__(self, writePriority, *args, **kwargs)
```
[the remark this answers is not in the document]
M4rtinK: Changed.

pykickstart/commands/ntp.py
```diff
@@ -0,0 +1,137 @@
...
+        self.op = self._getParser()
+        self.NTPList = kwargs.get("NTPList", [])
```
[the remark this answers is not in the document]
M4rtinK: Renamed.

pykickstart/commands/timesource.py
```diff
@@ -0,0 +1,146 @@
+#
+# Martin Kolman <mkolman@redhat.com>
```
bcl, line 2: Drop emails from copyright messages, we can use git blame to figure out who wrote things.

pykickstart/commands/timesource.py
```diff
@@ -0,0 +1,146 @@
...
+        self.nts = kwargs.get("nts", False)
+
+    def __eq__(self, y):
+        if not y:
+            return False
+
+        if self.ntp_server:
+            return self.ntp_server == y.ntp_server and self.nts == y.nts
+        elif self.ntp_pool:
+            return self.ntp_pool == y.ntp_pool and self.nts == y.nts
```
bcl, line 45: Need a return for the unexpected case or collapse it all into a single comparison, I think that's safe since the values all have reasonable empty defaults.

pykickstart/commands/timesource.py
```diff
@@ -0,0 +1,146 @@
...
+                        """)
+        op.add_argument("--ntp-disable", action="store_true",
+                        default=False, version=F33, help="""
+                        If specified, disable any NTP based time sync,
+                        both on target system as well as in installation
+                        environment.
+                        """)
+        op.add_argument("--nts", dest="nts", action="store_true",
+                        default=False, version=F33, help="""
+                        If specified, consider the server/pool to be
```
bcl, line 116: I think using `source` instead of the three `server/pool` makes it easier to read this.

tests/commands/timesource.py
```diff
@@ -0,0 +1,100 @@
+#
+# Martin Kolman <mkolman@redhat.com>
```
bcl, line 2: Same comment about emails in copyright messages

tests/commands/timesource.py
```diff
@@ -0,0 +1,100 @@
...
+# this program; if not, write to the Free Software Foundation, Inc., 51
+# Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.  Any Red Hat
+# trademarks that are incorporated in the source code or documentation are not
+# subject to the GNU General Public License and may only be used or replicated
+# with the express permission of Red Hat, Inc.
+#
+
+import unittest
+
+from pykickstart.errors import KickstartParseWarning
```
bcl, line 23: Remove unused imports: KickstartParseWarning, CommandSequenceTest, F33

tests/commands/timesource.py
```diff
@@ -0,0 +1,100 @@
...
+        self.assert_parse("timesource --ntp-pool=pool.example.com")
+        self.assert_parse("timesource --ntp-server=ntp.example.com --nts")
+        self.assert_parse("timesource --ntp-pool=pool.example.com --nts")
+        self.assert_parse("timesource --ntp-disable")
+
+        # fail
+        # server and pool can't be used at the same time
+        self.assert_parse_error("timesource --ntp-server=ntp.example.com --ntp-pool=pool.example.com")
+        self.assert_parse_error("timesource --ntp-server=ntp.example.com --ntp-pool=pool.example.com --nts")
+        # disable can(t be used with neither server or pool
```
bcl, line 61: Typo: `can't`

tests/commands/timesource.py
```diff
@@ -0,0 +1,100 @@
...
+        self.assert_parse_error("timesource --ntp-disable --ntp-pool=pool.example.com")
+        self.assert_parse_error("timesource --ntp-server=ntp.example.com --ntp-disable")
+        # --server requires an argument
```
bcl, line 64: Typo: `--ntp-server`

tests/commands/timesource.py
```diff
@@ -0,0 +1,100 @@
...
+        self.assert_parse_error("timesource --ntp-server")
+        self.assert_parse_error("timesource --ntp-server --nts")
+        # --pool requires int argument
```
bcl, line 67: Typo: `--ntp-pool`

pykickstart/commands/timesource.py
```diff
@@ -0,0 +1,146 @@
...
+
+class F33_Timesource(KickstartCommand):
+    removedKeywords = KickstartCommand.removedKeywords
+    removedAttrs = KickstartCommand.removedAttrs
+
+    def __init__(self, writePriority=0, *args, **kwargs):
+        KickstartCommand.__init__(self, writePriority, *args, **kwargs)
+        self.op = self._getParser()
+        self.timesource_list = kwargs.get("timesource_list", [])
+        self.exclusive_required_options = [("ntp_server", "--ntp_server"),
```
bcl, line 82: The arguments are using dash not underscore so these 3 should be `--ntp-server`, etc.

pykickstart/commands/timesource.py
```diff
@@ -0,0 +1,146 @@
...
+
+                        ``--ntp-server`` and ``--ntp-pool`` are mutually exclusive.
+                        """,
+                        metavar="")
+        op.add_argument("--ntp-pool", version=F33, help="""
+                        A single NTP pool.
+
+                        ``--ntp-pool`` and ``--ntp-server`` are mutually exclusive.
+                        """)
+        op.add_argument("--ntp-disable", action="store_true",
```
bcl, line 108: The commit message says `--disable` which is better if it should disable all time sources (assuming future addition of other non-ntp time sources). Or if it should only disable the NTP time source then `--ntp-disable` is better.

pykickstart/commands/timesource.py
VladimirSlavik: Maybe even "host"?

pykickstart/commands/ntp.py
M4rtinK: Added as `--ntp-disable`. :)

pykickstart/commands/timesource.py
M4rtinK: I've updated the commit message, that was indeed outdated. The `--ntp-disabled` option indeed should disable only NTP based time sources and their use, as Miroslav Lichvar, the NTP maintainer in Fedora mentioned to me that there are potentially other time sources we might want to support, so it makes sense to namespace these options like we do also with the `--ntp-server` and `--ntp-pool` options.

pykickstart/commands/timesource.py
M4rtinK: Indeed, looks like I got it mixed up somehow. Fixed. :)

tests/commands/timesource.py
M4rtinK: Fixed.

tests/commands/timesource.py
M4rtinK: Fixed.

tests/commands/timesource.py
M4rtinK: Fixed. :)

tests/commands/timesource.py
M4rtinK: Cleaned up. :)

tests/commands/timesource.py
M4rtinK: Removed.

pykickstart/commands/timesource.py
M4rtinK: Removed.

pykickstart/commands/timesource.py
M4rtinK: Looking into it now I think that given one of ntp_server/ntp_pool/ntp_disable is always required
and plain timesource is not valid, the else branch will be --ntp_disable, which should always be the same.

So fixed up accordingly.

pykickstart/commands/timesource.py
M4rtinK: I think "host" sounds good for the first one, "time source" for the second and third.

Fixed up accordingly, let me know what you think. :)

pykickstart/commands/ntp.py
M4rtinK: Yep, added in a separate commit - was rather easy to do as we recently deprecated an option of the timezone command (in Fedora 32) and this is very similar.

pykickstart/commands/ntp.py
```diff
@@ -0,0 +1,137 @@
...
+    removedKeywords = BaseData.removedKeywords
+    removedAttrs = BaseData.removedAttrs
+
+    def __init__(self, *args, **kwargs):
+        BaseData.__init__(self, *args, **kwargs)
+        self.server = kwargs.get("server", [])
+        self.pool = kwargs.get("pool", "")
+        self.nts = kwargs.get("nts", False)
+
+    def __eq__(self, y):
```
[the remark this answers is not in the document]
M4rtinK: I looked at the other commands that can be used multiple times and all had these data holders (`user`, `repo`, `partition`). I guess @bcl might know if these are still needed, or are just a historical artifact.

pykickstart/commands/ntp.py
```diff
@@ -0,0 +1,137 @@
...
+        return op
+
+    def parse(self, args):
+        nd = self.dataClass()   # pylint: disable=not-callable
+        ns = self.op.parse_args(args=args, lineno=self.lineno)
+        self.set_to_obj(ns, nd)
+        nd.lineno = self.lineno
+
+        # Check that just one of exclusive required options is specified
+        used_options = [opt for attr, opt in self.exclusive_required_options
```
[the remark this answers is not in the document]
M4rtinK: There is nothing like that unfortunately in the KSOptionParser class as far as I can tell. The same code is used in repo and url:

https://github.com/pykickstart/pykickstart/blob/master/pykickstart/commands/repo.py#L279
https://github.com/pykickstart/pykickstart/blob/master/pykickstart/commands/url.py#L189

This is effectively the third copy of it, so it indeed makes sense to de-duplicate it - @bcl any suggestions where to put it in via followup up PR (I think a PR adding a new command should ideally not do backend changes like this one)?

tests/commands/timezone.py
```diff
@@ -196,5 +196,26 @@ def runTest(self):
...
+
+class F33_TestCase(F32_TestCase):
+    command = "timezone"
+
+    def setUp(self):
+        super().setUp()
+        warnings.simplefilter("error", category=KickstartDeprecationWarning)
+
+    def runTest(self):
+        # As of Fedora 33 the --ntpservers and --nontp options are considered deprecated.
```
bcl, line 208: Shouldn't this also run the F32_TestCase.runTest? Otherwise we lose all those tests.

bcl:
Hmm, the commit message that GH is showing doesn't have --disabled changed for some reason.
I also just noticed a typo in the first word of the commit message -- `Witt` should be `With`  :)

pykickstart/commands/ntp.py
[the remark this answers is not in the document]
bcl: I don't quite understand the question :) How else would you handle multiple entries? The data holder is used to hold each one, and a list of them is returned by dataClass.

pykickstart/commands/ntp.py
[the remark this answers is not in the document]
bcl: I guess options.py would make sense, since it is dealing with options.

tests/commands/timezone.py
M4rtinK: Looking at the other tests cases for timezone, only the FC6 case runs the FC3 case - so I guess the others a re wrong ? Or I'm missing something obvious. :)

In any case, I'll add the `F32_TestCase.runTest` there and we can fix the other tests with separate PR if they are broken. :)

M4rtinK: Well, and now it makes the F32 test case fail, as it uses `--ntpservers`, which was deprecated. :D

Maybe that's the reason F32 does not call the older tests, as they would blow up in similar manner, as it also deprecates an option ? paging @VladimirSlavik as the author of the F32 test case :)

Why simple "add a new command" can't always be simple ? :D

pykickstart/commands/ntp.py
M4rtinK: Nope, copy paste artifact, fixed. :)

pykickstart/commands/ntp.py
```diff
@@ -0,0 +1,137 @@
...
+        self.nts = kwargs.get("nts", False)
+
+    def __eq__(self, y):
+        if not y:
+            return False
+
+        if self.server:
+            return self.server == y.server and self.nts == y.nts
+        elif self.pool:
+            return self.pool == y.pool and self.nts == y.nts
```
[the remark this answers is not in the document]
M4rtinK: Yep, should be fixed now with handling the third case (`--nts` option only).

pykickstart/commands/ntp.py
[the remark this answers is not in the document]
M4rtinK: > I guess options.py would make sense, since it is dealing with options.

OK, will do a followup PR to deduplicate this once this one is merged (so that we don't block implementation of the F33 NTS support change).

pykickstart/commands/timezone.py
```diff
@@ -420,3 +420,26 @@ def parse(self, args):
                           KickstartDeprecationWarning)
 
         return self
+
+
+class F33_Timezone(F32_Timezone):
+
+    def __init__(self, writePriority=0, *args, **kwargs):
+        F32_Timezone.__init__(self, writePriority, *args, **kwargs)
```
jkonecny12, line 428: What about using `super()` here?

pykickstart/commands/timezone.py
```diff
@@ -420,3 +420,26 @@ def parse(self, args):
...
+        self.op = self._getParser()
+
+    def parse(self, args):
+        F32_Timezone.parse(self, args)
```
jkonecny12, line 432: The same here.

tests/commands/timezone.py
```diff
@@ -196,5 +196,27 @@ def runTest(self):
...
+
+class F33_TestCase(F32_TestCase):
+    command = "timezone"
+
+    def setUp(self):
+        super().setUp()
+        warnings.simplefilter("error", category=KickstartDeprecationWarning)
+
+    def runTest(self):
+        F32_TestCase.runTest(self)
```
jkonecny12, line 208: The same here.

pykickstart/commands/timesource.py
```diff
@@ -0,0 +1,149 @@
...
+        return retval
+
+    def _getParser(self):
+        op = KSOptionParser(prog="timesource", description="""
+                            Configures a timesource.""",
+                            version=F33)
+        op.add_argument("--ntp-server", version=F33, help="""
+                        A single NTP server.
+
+                        ``--ntp-server`` and ``--ntp-pool`` are mutually exclusive.
```
poncovka, line 103: I guess that they should be also mutually exclusive with `--ntp-disable`.

tests/commands/timezone.py
```diff
@@ -196,5 +196,27 @@ def runTest(self):
                                 KickstartDeprecationWarning,
                                 "The option --isUtc will be deprecated in future releases")
 
+
+class F33_TestCase(F32_TestCase):
+    command = "timezone"
+
+    def setUp(self):
+        super().setUp()
+        warnings.simplefilter("error", category=KickstartDeprecationWarning)
```
poncovka, line 205: I don't think we should transform warnings into exceptions, because it can break tests for older versions. This is happening to F32 and F33 timezone tests. Please, remove this line from both test cases:

`warnings.simplefilter("error", category=KickstartDeprecationWarning)`

The warnings can be simply tested with:


```
        with self.assertWarns(KickstartDeprecationWarning):
            self.assert_parse("timezone --isUtc Europe/Bratislava")

```

tests/commands/timesource.py
```diff
@@ -0,0 +1,96 @@
...
+            data2 = F33_TimesourceData()
+            setattr(data1, attr, '')
+            setattr(data2, attr, 'test')
+            self.assertNotEqual(data1, data2)
+            self.assertNotEqual(data2, data1)
+            setattr(data1, attr, '')
+            setattr(data2, attr, '')
+
+        # pass
+        self.assert_parse("timesource --ntp-server=ntp.example.com")
```
poncovka, line 47: It would be nice to have at least one test with multiple sources.

pykickstart/commands/ntp.py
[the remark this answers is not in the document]
poncovka: I was asking specifically about the `__eq__` and `__ne__` methods. The other data holders use these methods to check for duplicates. This doesn't apply to timesources.

pykickstart/commands/timezone.py
M4rtinK: No other commands use super() at the moment, if we want to change this we should do a refactor globally via separate PR (provided @bcl is fine with that).

pykickstart/commands/timezone.py
M4rtinK: See above.

tests/commands/timezone.py
M4rtinK: See above.

pykickstart/commands/timesource.py
M4rtinK: Fixed.

tests/commands/timezone.py
M4rtinK: Yep, that looks much better than the old code from the F32 test case I have reused. :)

tests/commands/timesource.py
M4rtinK: Added.

pykickstart/commands/ntp.py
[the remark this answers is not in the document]
bcl: Ah, ok. I don't think it hurts to have them but it should be ok to leave them out if they aren't used.

pykickstart/commands/timezone.py
jkonecny12: OK.

tests/commands/timezone.py
M4rtinK: This should be resolved with the latest bunch of fixes.

tests/commands/timesource.py
M4rtinK: Added.

M4rtinK:
@bcl Thanks! :)

pykickstart/commands/timesource.py
```diff
@@ -0,0 +1,148 @@
...
+
+        if self.ntp_server:
+            return self.ntp_server == y.ntp_server and self.nts == y.nts
+        elif self.ntp_pool:
+            return self.ntp_pool == y.ntp_pool and self.nts == y.nts
+        else:
+            # given that one of ntp_server/ntp_pool/ntp_disable is required
+            # this will be --ntp_disable & such timesource invocations will
+            # always be the same by definition
+            return True
```
VladimirSlavik, line 48: I don't see how this could always be true? The assumption about `self` being `--ntp_disable` seems correct to me, but not the one about `y` being the same...

tests/commands/timezone.py
VladimirSlavik: Sorry, missed this :(

poncovka:
Hi @bcl, could you merge this PR and release a new version of pykickstart, please? I have tested it with https://github.com/rhinstaller/anaconda/pull/2704 and it seems to work fine.

bcl:
There's a couple pylint warnings about the test code, I'll fix those in a new commit before doing the build.

PR #325, opened 2020-07-30
Remove OrderedSet usage

bcl:


jcpunk:
When the final `anaconda-ks.cfg` is written out, I'll confess I like having a sorted list of packages.  It makes doing a quick `diff` against two different kickstarts much cleaner and, to my mind, improves readability.  But so long as the parsed output is sorted, I don't much care what pykickstart does internally.

bcl:
> When the final `anaconda-ks.cfg` is written out, I'll confess I like having a sorted list of packages. It makes doing a quick `diff` against two different kickstarts much cleaner and, to my mind, improves readability. But so long as the parsed output is sorted, I don't much care what pykickstart does internally.

That's what I'm hoping for :) Also, if you flatten a kickstart, like Fedora does for the live iso, the output will all be sorted as well.

PR #330, opened 2020-08-07
Remove unneeded refs to deps

CMeza99:
* cleaned up references to `nose`
* removed explicitly calling out Python3 as a requirement since py2 is no longer supported


bcl:
Tried to use the github conflict editor thing, but it wasn't happy (maybe because it was trying to make a merge commit) so I manually picked these over to master.

PR #333, opened 2020-08-27
Remove harddrive --biospart support

jkonecny12:
It's not supported by Anaconda for a long time. All the Anaconda code was using `biospart` attribute was that it's not yet supported and this code was in Anaconda from 2012. Let's remove this completely and make everything simpler.

See https://github.com/rhinstaller/anaconda/pull/2810

-------------------------
Is there a way how to avoid changing old test case when removing arguments? Did I missed something?

### DIFF
--- a/tests/commands/harddrive.py
+++ b/tests/commands/harddrive.py
@@ -20,30 +20,37 @@
 
 import unittest
 from tests.baseclass import CommandTest
-from pykickstart.commands.harddrive import FC3_HardDrive
+from pykickstart.commands.harddrive import FC3_HardDrive, F33_HardDrive
+
+
+def attribute_test(test_class, data1, data2, compare_attrs):
+    # additional test coverage
+    test_class.assertEqual(data1.__str__(), '')
+
+    # test that new objects are always equal
+    test_class.assertEqual(data1, data2)
+    test_class.assertNotEqual(data1, None)
+
+    # test for objects difference
+    for atr in compare_attrs:
+        setattr(data1, atr, '')
+        setattr(data2, atr, 'test')
+        # objects that differ in only one attribute
+        # are not equal
+        test_class.assertNotEqual(data1, data2)
+        test_class.assertNotEqual(data2, data1)
+        setattr(data1, atr, '')
+        setattr(data2, atr, '')
+
 
 class HardDrive_TestCase(unittest.TestCase):
     def runTest(self):
         data1 = FC3_HardDrive()
         data2 = FC3_HardDrive()
 
-        # additional test coverage
-        self.assertEqual(data1.__str__(), '')
-
-        # test that new objects are always equal
-        self.assertEqual(data1, data2)
-        self.assertNotEqual(data1, None)
-
-        # test for objects difference
-        for atr in ['biospart', 'partition', 'dir']:
-            setattr(data1, atr, '')
-            setattr(data2, atr, 'test')
-            # objects that differ in only one attribute
-            # are not equal
-            self.assertNotEqual(data1, data2)
-            self.assertNotEqual(data2, data1)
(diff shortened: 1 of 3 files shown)


tests/commands/harddrive.py
```diff
@@ -75,5 +70,17 @@ def runTest(self):
         # unknown option
         self.assert_parse_error("harddrive --unknown=value")
 
+
+class F33_TestCase(FC3_TestCase):
```
bcl, line 74: If you don't base it on the the FC3 testcase then you won't have to change the FC3 code. You'll have to duplicate the rest of the tests in here though.

Previous versions of kickstart need to continue to support it (eg. when passing --version=F27), so we shouldn't change the older tests.

jkonecny12: I leaved the old tests untouched and make these independent. Hope I did it correctly. :)

jkonecny12:
**UPDATED:**
* Do not modify the old tests.

tests/commands/harddrive.py
jkonecny12: I've also tried to extend the "compare" test. If there is other way how to do this correctly I'll change it appropriately.

tests/commands/harddrive.py
```diff
@@ -20,30 +20,35 @@
...
+def attribute_test(data1, data2, test_class):
+    # additional test coverage
+    test_class.assertEqual(data1.__str__(), '')
+
+    # test that new objects are always equal
+    test_class.assertEqual(data1, data2)
+    test_class.assertNotEqual(data1, None)
+
+    # test for objects difference
+    for atr in ['biospart', 'partition', 'dir']:
```
bcl, line 35: The attributes also need to be passed in, that's why the travis test is failing.

tests/commands/harddrive.py
```diff
@@ -75,5 +80,36 @@ def runTest(self):
...
+                        self.assert_parse("harddrive --dir=/install --partition=sda2"))
+
+        # fail
+        # Ensure these options have been removed.
+        self.assert_removed("harddrive", "--biospart")
+        # required option --dir missing
+        self.assert_parse_error("harddrive")
+        # required --dir argument missing
+        self.assert_parse_error("harddrive --dir")
+        # missing --biospart or --partition option
```
bcl, line 107: No more biospart :)

tests/commands/harddrive.py
```diff
@@ -75,5 +80,36 @@ def runTest(self):
...
+        self.assert_parse_error("harddrive --dir=/install")
+        # both --biospart and --partition specified
```
bcl, line 109: No more biospart

tests/commands/harddrive.py
```diff
@@ -75,5 +80,36 @@ def runTest(self):
...
+        self.assert_parse_error("harddrive --dir=/install --partition")
+        # unknown option
+        self.assert_parse_error("harddrive --unknown=value")
+
```
bcl, line 113: Add checks for `harddrive --partition` and `harddrive --partition=/install` since both `--partition` and `--dir` are now required parameters.

tests/commands/harddrive.py
jkonecny12: Fixed.

tests/commands/harddrive.py
jkonecny12: Fixed.

tests/commands/harddrive.py
jkonecny12: Fixed.

tests/commands/harddrive.py
jkonecny12: Fixed if I understand it correctly that the second should be ``... --dir=/install``.

Issue #339, opened 2020-09-11
Error when initializing `KickstartCompleter` object


zuh0:
Hello,

When launching `ksshell` I get a Python stack trace because of an `AttributeError`.

Expected behaviour:

Launching `ksshell`.

Observed behavior:

A stack trace:

```
Traceback (most recent call last):
  File "/home/zuh0/src/pykickstart/venv/bin/ksshell", line 240, in <module>
    main()
  File "/home/zuh0/src/pykickstart/venv/bin/ksshell", line 187, in main
    readline.set_completer(KickstartCompleter(kshandler, internalCommands).complete)
  File "/home/zuh0/src/pykickstart/venv/bin/ksshell", line 84, in __init__
    self._add_command(cStr, cObj)
  File "/home/zuh0/src/pykickstart/venv/bin/ksshell", line 98, in _add_command
    for opt in cObj.op.option_list:
AttributeError: 'KSOptionParser' object has no attribute 'option_list'
```

Steps to reproduce:

```
$ git clone https://github.com/pykickstart/pykickstart
$ cd pykickstart
$ python3 -m venv venv
$ . venv/bin/activate
$ python3 -m pip install -r requirements.txt
$ ksshell
```

Python version:  3.8.5
Linux distribution: Fedora 32

I have also checked in the official Fedora 32 repositories and the packaged version is also broken in the same way.

bcl:
Thanks, I've filed a bugzilla bug for this - https://bugzilla.redhat.com/show_bug.cgi?id=1878242

bcl:
Fixed by PR #340

PR #340, opened 2020-09-11
ksshell: Fix traceback and add tests

bcl:
This fixes the traceback with ksshell and adds a simple test to make
sure it is minimally functional.

Resolves: rhbz#1878242


bcl:
Fixes #339

jkonecny12 (approved):
Looks good to me ;)

PR #343, opened 2020-10-22
Deprecate autostep in RHEL8

VladimirSlavik:
It doesn't really work anymore.

### DIFF
--- a/pykickstart/commands/autostep.py
+++ b/pykickstart/commands/autostep.py
@@ -17,8 +17,8 @@
 # subject to the GNU General Public License and may only be used or replicated
 # with the express permission of Red Hat, Inc.
 #
-from pykickstart.version import FC3
-from pykickstart.base import KickstartCommand
+from pykickstart.version import FC3, RHEL8, versionToLongString
+from pykickstart.base import KickstartCommand, DeprecatedCommand
 from pykickstart.options import KSOptionParser
 
 class FC3_AutoStep(KickstartCommand):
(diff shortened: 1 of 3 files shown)


VladimirSlavik:
Only as draft because it's not clear yet if we will do this.

tests/commands/autostep.py
```diff
@@ -19,7 +19,8 @@
 #
 import unittest
 from tests.baseclass import CommandTest
-from pykickstart.commands.autostep import FC3_AutoStep
+from pykickstart.commands.autostep import FC3_AutoStep, RHEL8_AutoStep
```
VladimirSlavik, line 22: No need to import `RHEL8_AutoStep`

pykickstart/commands/autostep.py
```diff
@@ -17,8 +17,8 @@
 # subject to the GNU General Public License and may only be used or replicated
 # with the express permission of Red Hat, Inc.
 #
-from pykickstart.version import FC3
-from pykickstart.base import KickstartCommand
+from pykickstart.version import FC3, RHEL8
```
VladimirSlavik, line 20: Import `versionToLongString` here.

VladimirSlavik (commented):
After landing the F34 changes in #344, some errors in this are apparent.

VladimirSlavik:
It seems to me that Pylint got upgraded in Travis (with new Ubuntu that they warned of), the errors don't look to me like "mine".

VladimirSlavik:
@bcl this is blocked by #371, right?

bcl:
Not really, I can run the tests locally before merging. But it does need a rhbz with acks.

bcl (changes requested):
Rebase this PR on the current rhel8-branch so that the new GH Action based tests run.

The Commit message also needs a Resolves: rhbz#1895951 line added to it.

VladimirSlavik:
Thanks!

PR #344, opened 2020-10-22
Deprecate the autostep command

VladimirSlavik:
It doesn't really work in Anaconda anymore.

### DIFF
--- a/tests/commands/autostep.py
+++ b/tests/commands/autostep.py
@@ -20,6 +20,7 @@
 import unittest
 from tests.baseclass import CommandTest
 from pykickstart.commands.autostep import FC3_AutoStep
+from pykickstart.base import DeprecatedCommand
 
 class FC3_TestCase(CommandTest):
     command = "autostep"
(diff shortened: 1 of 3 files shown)


poncovka (commented):
Looks good to me. Thanks!

tests/commands/autostep.py
```diff
@@ -19,7 +19,8 @@
 #
 import unittest
 from tests.baseclass import CommandTest
-from pykickstart.commands.autostep import FC3_AutoStep
+from pykickstart.commands.autostep import FC3_AutoStep, F34_AutoStep
```
bcl, line 22: Importing `F34_AutoStep` isn't needed, and makes pylint warn about an unused import. Fix that and it looks fine to me.

bcl (changes requested):

tests/commands/autostep.py
VladimirSlavik: Thanks, done.

PR #345, opened 2020-11-02
test: Update Dockerfile to use CentOS 8

bcl:
Fedora 29 is EOL and we now have CentOS available to use for running the
tests. This primarily effects the Travis tests run by the github pull
requests.

Use packages for as much of the test requirements as possible. This
means adding the EPEL8 repository and trimming the pip3 install to only
2 packages, pocketlint and Sphinx.

Resolves: rhbz#xxxxxx


Dockerfile.test
```diff
@@ -1,8 +1,10 @@
...
 
 # Add test dependencies
-COPY requirements.txt /requirements.txt
-RUN dnf -y install make git gettext python3-pip python3-sphinx which
-RUN pip-3 install -r requirements.txt
+COPY test-packages .
+RUN dnf -y install $(cat test-packages)
+
+# pocketlint to modify pylint behavior, and Sphinx because it isn't in EPEL8
+RUN pip-3 install pocketlint 'Sphinx == 1.7.6'
```
Conan-Kudo, line 10: Sphinx is present in the PowerTools repo, so you can enable that to install `python3-sphinx` from CentOS.

Conan-Kudo (changes requested):

Dockerfile.test
bcl: Thanks!

PR #346, opened 2020-11-06
Check if UID and GID values are valid

VladimirSlavik:
This makes wrong values of GID and UID fail. Wrong means not a number, and outside the allowed range.

(Draft so far, because this needs more discussion. I just wanted to see how this would look like.)


pykickstart/options.py
```diff
@@ -74,6 +74,16 @@ def mountpoint(value):
 
     return value
 
+def uid_gid(value):
+    try:
+        number = int(value)
+        if number > 2**32-1 or number < 1:
+            raise ArgumentTypeError(_("UID or GID value out of permitted range: %r") % value)
+        return number
+    except (ValueError, TypeError):
```
VladimirSlavik, line 83: Aside from the larger implications of this, I am interested in this bit - Do I need to allow None here?

pykickstart/options.py
```diff
@@ -74,6 +74,16 @@ def mountpoint(value):
 
     return value
 
+def uid_gid(value):
+    try:
+        number = int(value)
+        if number > 2**32-1 or number < 1:
```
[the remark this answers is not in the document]
poncovka: Btw. you don't have to keep this check in the try-except block, you can move it behind it.

pykickstart/options.py
poncovka: I think that `uid_gid` shoudn't be called when `--gid` (or `--uid`) is not specified in the kickstart file. You can check the documentation of the argument parser.

poncovka (commented):
Otherwise, it looks good to me. Thanks!

VladimirSlavik:
After testing together with https://github.com/rhinstaller/anaconda/pull/2982, it seems a too large UID is not caught with this change. Needs more work...

bcl:
I'm not sure that checking limits belongs in kickstart. I'd prefer to allow the underlying system to fail and pass that failure on to the user instead of hard-coding limits.

What practical problem is this trying to solve? In general I don't think I've ever seen someone trying to use a uid/gid that's too big.

OTOH it's probably ok to do a digit check to prevent passing in names on accident.

PR #348, opened 2020-11-09
Remove the bootloader option --upgrade

VladimirSlavik:
It has been deprecated since F29.


poncovka (commented):
Looks good to me. Thanks!

PR #352, opened 2020-11-11
Deprecate the method command

VladimirSlavik:
Another no longer used command.

### DIFF
--- a/tests/commands/method.py
+++ b/tests/commands/method.py
@@ -21,6 +21,7 @@
 import unittest
 import copy
 from tests.baseclass import CommandTest
+from pykickstart.base import DeprecatedCommand
 
 class FC3_Proxy_TestCase(CommandTest):
     command = "method"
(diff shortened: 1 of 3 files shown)


poncovka:
This pull request is more like a proposal. Anaconda no longer uses the `method` command. We use the related kickstart commands directly and we don't need to change the installation method on the pykickstart level anymore. I am not able to tell whether the command is useful for someone else.

jkonecny12 (approved):
Die, die, die! No sorry it wasn't professional but this caused so many bugs so it would be great to get rid of this.
I just wonder it's a bit core API. Isn't this used by someone else than Anaconda team? On the other hand it's 'just' deprecation so it could cause people to stand up and tell.

VladimirSlavik:
It's the pykickstart-style deprecation, so really it means "no longer works".

bcl:
> I am not able to tell whether the command is useful for someone else.

IIRC it used to be used by Beaker when installing systems, and I cannot be 100% sure they have stopped using it -- but if Anaconda no longer supports it then removing it should be fine.

tests/commands/method.py
```diff
@@ -186,7 +187,8 @@ def runTest(self):
         self.assert_parse("cdrom", "cdrom\n")
 
         # harddrive
-        self.assert_parse("harddrive --dir=/install --biospart=part", "harddrive --dir=/install --biospart=part\n")
```
bcl, line 189: This appears to be unrelated to the method deprecation.

bcl (changes requested):

tests/commands/method.py
VladimirSlavik: Well, yes and no. When `--biospart` was removed for F33, the `harddrive` test was updated, but not the `method` one, which stayed stuck at the older command version and apparently all was well. Now that is no longer the case.

bcl: Oh! I see now, thanks!

bcl (approved):
Looks good!

Issue #350, opened 2020-11-10
Clean up Questions Surrounding Deprecation


bcl:
There's some question about the correct behavior of pykickstart's deprecated commands.
It appears that in some cases deprecation has removed/disabled parsing of the command immediately.
It should instead continue to be parsed, but include a deprecation warning.

Need to clean this up and clearly document how to deprecate commands and arguments.

VladimirSlavik:
Thank you!

bcl:
Fixed by #356

PR #356, opened 2020-11-20
Document how deprecation currently works

bcl:


bcl:
This will fix #350

CONTRIBUTING
```diff
@@ -121,3 +121,287 @@ Bad Contributions
       something that kickstart needs to capture and provide to programs using
       it.
 
+
+KICKSTART DEPRECATION
+
+Over time commands and arguments change. New ones are added and old ones are
+eventually removed.  Kickstart's use of the term 'deprecation' can be a bit
+confusing.  At the time of this writing, version 3.31, a deprecated command
```
jkonecny12, line 129: I would remove ``Kickstart's use of the term 'deprecation' can be a bit confusing.`` it's mentioned in the next paragraph too.

jkonecny12 (approved):
Looks good to me otherwise. Thanks for this great documentation improvement!

VladimirSlavik (approved):
Thank you, this is great! A couple of points though:

* You mention that a custom deprecation message can be added, but *is not ideal*. There's some misunderstanding here. I am told that for RHEL compatibility, there needs to be in successive major releases first a warning (custom message), then loss of functionality (deprecated), and only then an error (removed). And after all, it seems logical to have some notification that things will go away, while they still work - that's how it happens elsewhere, see eg. Python warnings about `imp`. Anyway, my takeaway was that this state is needed.

* My inner documentation writer cringed at no markup for all of the identifiers, paths etc. That does not detract from the information value, though, so feel free to ignore.

pykickstart/options.py
```diff
@@ -257,6 +257,18 @@ def remove_argument(self, arg, **kwargs):
             self._remove_action(candidate)
             self._option_string_actions.pop(arg)
 
+    def deprecate_argument(self, *args, deprecated):
```
poncovka, line 260: This method is not mentioned in the CONTRIBUTING file.

poncovka: Shouldn't it be similar to `remove_argument` and accept `arg` instead of `*args`?

poncovka (approved):
Otherwise, it looks great. Thank you!

pykickstart/options.py
VladimirSlavik: This might work better as a separate followup PR which would add the new function, use it, and adjust the docs. Without actual usage it's a guess at best.

bcl: Yeah, that's my plan. I included it here for discussion and so I didn't lose track of it :)

bcl: > Shouldn't it be similar to `remove_argument` and accept `arg` instead of `*args`?

`*args` makes it easier to deprecate multiple arguments at the same time. Might be useful to also change remove_argument.

bcl:
> Thank you, this is great! A couple of points though:
> 
>     * You mention that a custom deprecation message can be added, but _is not ideal_. There's some misunderstanding here. I am told that for RHEL compatibility, there needs to be in successive major releases first a warning (custom message), then loss of functionality (deprecated), and only then an error (removed). And after all, it seems logical to have some notification that things will go away, while they still work - that's how it happens elsewhere, see eg. Python warnings about `imp`. Anyway, my takeaway was that this state is needed.

That's where the mismatch happens. Deprecation means a warning, not loss of functionality. Removal of it in Anaconda is a separate state, as is removal in pykickstart. If we do those things at separate times then I think Anaconda should be the one to tell the user that it is no longer implemented.

> 
>     * My inner documentation writer cringed at no markup for all of the identifiers, paths etc. That does not detract from the information value, though, so feel free to ignore.

This file isn't markdown, so no markup :)

VladimirSlavik:
Oh. I see where you're coming from - yes, that makes sense. But then deprecation should not be a pykickstart thing either?

bcl:
> Oh. I see where you're coming from - yes, that makes sense. But then deprecation should not be a pykickstart thing either?

I suppose you could make that argument, but having deprecation should up with ksvalidator is a good prompt for them to change their kickstart without running into installation problems.

PR #365, opened 2020-12-04
One way we could document command removal

bcl:


pykickstart/commands/interactive.py
```diff
@@ -56,3 +56,9 @@ def _getParser(self):
         op = FC3_Interactive._getParser(self)
         op.description += "\n\n.. deprecated:: %s" % versionToLongString(F14)
         return op
+
+class F15_Interactive(RemovedCommand):
+    def _getParser(self):
+        op = F14_Interactive._getParser(self)
+        op.description += "\n\n.. versionremoved:: %s" % versionToLongString(F15)
```
jkonecny12, line 63: I wonder why this have to be specified manually and can't be implemented by the ``RemovedCommand``.

jkonecny12 (commented):
Looks great to me. What do you think @VladimirSlavik ?

pykickstart/commands/interactive.py
VladimirSlavik: Seems like a job for a macro :P (joking)

VladimirSlavik:
I was thinking more along the lines of keeping the pykickstart code proper as is. Instead, the docs generation could find when the command is last available. That would automatically fill in all the blanks, just as they are right now. Something similar to finding the highest command version, except in reverse, finding whether the command is handled. What do you think @bcl?

VladimirSlavik:
I tried to modify the docs generating script. I was able to work around a number of things, and it sort of works. Here's the code: https://github.com/VladimirSlavik/pykickstart/commit/e808d9ebddf30ffb65d17c99159a3736525b48ad

The remaining problems I have are:

a) Naively comparing version constants can yield weirdness since RHEL and Fedora are "multiplexed", and finding next release is needed since that's where the command is NOT seen.

b) The removal string needs to go somewhere higher than at the end of command output. Could be a warning even before syntax details, or some mechanism is needed to inject it into the parser help string.

pykickstart/commands/interactive.py
bcl: The class doesn't know how or where to get the previous _getParser from. I tried playing with some mixins a bit and it just ended up being a mess and not working  completely.

Doing it this way also matches how deprecation is handled so it's easier for infrequent contributors to understand.

bcl:
> But I was thinking more along the lines of keeping the pykickstart code proper as is. Instead, the docs generation could find when the command is last available.

I considered that, but really don't want to load up the document generator with a bunch of code.
Another benefit to doing it explicitly is that the code clearly reflects when it was removed, so there is no need to look for the docs, run ksverdiff, or manually compare the handler lists.

VladimirSlavik:
I fully agree with keeping it in code part as something desirable. The main con of this PR is that then *someone* would have to go through all the commands and mark them as removed...

Though, that might be a nice way to relax :D

bcl:
> The main con of this PR is that then _someone_ would have to go through all the commands and mark them as removed...

We don't have to add them right away, and really there aren't that many of them. According to ksverdiff from FC3 to F34:

interactive langsupport lilo lilocheck monitor mouse upgrade

Are the only ones.

pykickstart/options.py
```diff
@@ -149,6 +156,9 @@ def __init__(self, *args, **kwargs):
         if addVersion:
             kwargs['description'] = "\n.. versionadded:: %s\n\n%s" % (version,
                                                                     kwargs['description'])
+        if removed:
+            kwargs['description'] = "\n.. versionremoved:: %s\n\n%s" % (removed,
```
poncovka, line 160: This is not used in the example. Is there really a use case for it?

poncovka (approved):
I agree with @VladimirSlavik that it would be great to handle removed commands automatically, but this solution is good enough.

VladimirSlavik (approved):
OK then, let's have it! Thank you!

pykickstart/options.py
bcl: Oops, we already have the `KSOptionParser.remove_argument()` function that handles doing that. I'll drop that part, thanks!

PR #371, opened 2021-01-13
Backport GH Actions to rhel8-branch

bcl:


VladimirSlavik:
Looks like you're getting the same pylint-only errors as #343, so I think it's good?

.github/workflows/tests.yml
```diff
@@ -0,0 +1,44 @@
...
+      max-parallel: 5
+      matrix:
+        python-version: [3.6, 3.7]
+        include:
+          - python-version: 3.6
+            tox-env: py36
+          - python-version: 3.7
+            tox-env: py37
+    steps:
+      - name: "Clone Repository"
```
jkonecny12, line 28: You should also rebase the PR on top of the target branch. Something like:
https://github.com/rhinstaller/anaconda/blob/master/.github/workflows/validate-rhel-8.yml#L70

However, you need to get target branch for the given PR which should be doable by https://docs.github.com/en/free-pro-team@latest/actions/reference/environment-variables#default-environment-variables .

jkonecny12 (commented):
Looks good to me otherwise.

.github/workflows/tests.yml
bcl: I don't think so. This runs in pull_request, and the clone step gets the right code. IIRC it was pull_request_target where you have to rebase things. https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#pull_request

jkonecny12: No, with `pull_request_target` you will have the same code as `pull_request`. However, this is not what I meant. When you look on [checkout action](https://github.com/actions/checkout) you will find:

> Only a single commit is fetched by default, for the ref/SHA that triggered the workflow. 

What triggered this code is SHA of the PR not rebase of the PR on the target branch.

The point is, that you are testing exact code which you have on PR. However, the target HEAD of the PR is moving, so you are not testing what will be there after the merge will happen but before the merge. This could be problematic and could miss problems especially when the PR in question is old.

bcl: Ah! I see. That's not usually a problem for me, and when it is I just rebase the PR and force-push it to run it against the new HEAD.

I'm not sure if I like the idea of doing that automatically -- when I write the PR I write it against a specific branching point, if the GH test then tests it against a different point that seems unexpected to me. I'll have to think about that.

bcl:
Note that this needs a rhbz in order to be merged. I'll open one today and add it to the commits.

PR #372, opened 2021-01-15
Conditionally BuildRequire coverage and sphinx for runtests

bcl:
They aren't needed for the build unless runtests has been selected (it
is disabled by default).

Resolves: rhbz#1916735


bcl:
The dist-git change looks like this: https://src.fedoraproject.org/rpms/pykickstart/pull-request/3

PR #373, opened 2021-02-15
Add F34_MountData that enforces compression on btrfs (#1928857)

michel-slm:
On F34, we want to ensure that Btrfs filesystems are mounted with
transparent compression enabled; the compression algorithm and
level decided on was `compress=zstd:1`.

Ensure `MountData` always return this unless a different compression
setting is selected.

Signed-off-by: Michel Alexandre Salim <michel@michel-slm.name>

---
This change is [Reviewable](https://reviewable.io/reviews/pykickstart/pykickstart/373)

### DIFF
--- a/pykickstart/commands/mount.py
+++ b/pykickstart/commands/mount.py
@@ -25,6 +25,7 @@ from pykickstart.version import F27
 
 from pykickstart.i18n import _
 
+
 class F27_MountData(BaseData):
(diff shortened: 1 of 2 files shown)


bcl:
Kickstart isn't the right place to add filesystem specific changes like this. It should be added to Anaconda's kickstart handling code.

PR #374, opened 2021-03-12
rootpw: add option '--sshlogin'

kkang-wr:
Add option '--sshlogin' for command rootpw. It could remember and redo
the selection of option "Allow root SSH login with password" of anaconda
installler.

Signed-off-by: Kai Kang <kai.kang@windriver.com>

### DIFF
--- a/pykickstart/commands/rootpw.py
+++ b/pykickstart/commands/rootpw.py
@@ -17,7 +17,7 @@
 # subject to the GNU General Public License and may only be used or replicated
 # with the express permission of Red Hat, Inc.
 #
-from pykickstart.version import FC3, F8
+from pykickstart.version import FC3, F8, F34
 from pykickstart.base import KickstartCommand
 from pykickstart.errors import KickstartParseError
 from pykickstart.options import KSOptionParser
@@ -146,3 +146,29 @@ class F18_RootPw(F8_RootPw):
 
         self.set_to_self(ns)
         return self
+
+class F34_RootPw(F18_RootPw):
+    removedKeywords = F18_RootPw.removedKeywords
+    removedAttrs = F18_RootPw.removedAttrs
+
+    def __init__(self, writePriority=0, *args, **kwargs):
+        F18_RootPw.__init__(self, writePriority, *args, **kwargs)
+        self.sshlogin = kwargs.get("sshlogin", False)
+
+    def __str__(self):
+        retval = F18_RootPw.__str__(self)
+
+        if self.sshlogin:
+            retval = retval.rstrip('\n')
+            retval += " --sshlogin\n"
+
+        return retval
+
+    def _getParser(self):
+        op = F18_RootPw._getParser(self)
+        op.add_argument("--sshlogin", action="store_true", default=False,
+                        version=F34, help="""
+                        If this is present, it sets sshd configuration to allow
+                        root account log in via ssh. If the root account is
+                        locked, it can not log in via ssh either.""")
+        return op
--- a/pykickstart/handlers/f34.py
+++ b/pykickstart/handlers/f34.py
@@ -76,7 +76,7 @@ class F34Handler(BaseHandler):
         "repo": commands.repo.F33_Repo,
         "reqpart": commands.reqpart.F23_ReqPart,
         "rescue": commands.rescue.F10_Rescue,
-        "rootpw": commands.rootpw.F18_RootPw,
+        "rootpw": commands.rootpw.F34_RootPw,
         "selinux": commands.selinux.FC3_SELinux,
         "services": commands.services.FC6_Services,
         "shutdown": commands.reboot.F23_Reboot,


kkang-wr:
Hi,

I do not understand what is going wrong with coverage/coveralls test. Could anyone help me to figure out what should I do next to fix the issue, please?

Thanks a lot.

Kai

VladimirSlavik:
Hi @kkang-wr! I'm not the maintainer here, but hopefully this is enough to go on for now:

Personally I'd say, wait for the anaconda discussion, then come back to this. I'm new, but I think there must be some history if such an obvious feature is missing.

Anyway, for pykickstart, you have to add tests. The super-high coverage could not exist without tests for all the code, so that's the expectation anyway. Write a test for the `F34_` variant/version of the command, here: https://github.com/pykickstart/pykickstart/blob/master/tests/commands/rootpw.py

As to the original question, the coverage simply fails if it goes down. Even with a test for the new code, the coverage could be going down by some 0.00001% or so and "fail", but that's not a problem anymore. (I think it counts also some "meta" stuff like declarations, etc. so if the logic/meta ratio is low, it can go down even with full testing.)

Code looks good to me in terms of working. Actual testing with real data is another matter of course.

I think there are only two things with potential to request change:

1. Option name. I think the trend is to go `--words-with-dashes`.
2. Whether to use  a "add to enable" option like `--blah` (as you did), or an option that can go also `--blah=yes|no`. Personally I think this is fine as-is.

bcl:
As an overall philosophy I'm opposed to allowing password logins over ssh. If you can set a password in the kickstart you can just as easily set a ssh key using the `sshkey` command. But if Anaconda wants to add this then I guess I will accept it.

It looks like there is an associated Anaconda PR here - https://github.com/rhinstaller/anaconda/pull/3240

I think `--sshlogin` is ok, pykickstart doesn't use any yes/no options, and the default everywhere I looked is to deny root ssh password logins by default so an option to set it to no wouldn't be useful.

It does need tests, make sure to test in combination with the other options. I'm slightly unsure if `--lock` and `--sshlogin` make sense together, but we can always allow Anaconda to enforce restrictions on that.

kkang-wr:
@VladimirSlavik @bcl,

Thanks for your detailed replies.

The original requirement of this patch is from anaconda. The root password setting page has an checkbox option  'Allow root SSH login with password'. But it does not write anything to ks file when it is checked.

And I am not sure about `--lock` and `--sshlogin` either. Because when checkbox 'Lock root account' is checked, it does not disable the root ssh login checkbox. That why I did not check option `--lock` when add `--sshlogin`.

I'll add test case and rebase on branch F25.

Regards,
Kai

M4rtinK:
I don't think it is a good idea to add an option to enable password based root login. It's really not good practice to enable password based SSH root login on a machine as the attacker only needs to guess a password for root and then gets full access to the machine. For a user account the attacker needs to ques both the username and password and might only get to a non-admin user, making such an attack much harder and less worthwhile.

This was the reasoning for the OpenSSH project [disabling password logins for root back in 2015](http://www.openssh.com/txt/release-7.0). Fedora patched this out temporarily but in 2019 it was decided to [drop this downstream patch and respect the upstream behavior](https://fedoraproject.org/wiki/Changes/DisableRootPasswordLoginInSshd) of not allowing password based root login over SSH by default. Anaconda accommodated this change by adding and override checkbox in the root password GUI to make the transition easier for users still needing SSH login via root during the transition period.

While there is currently no set deadline for removing the checkbox from the GUI, it is still considered a temporary thing helping users during the transition to ideally there being no use cases requiring password based root login in the future. The option will most likely be dropped in the longer term, when it is considered no longer necessary - of course with a proper heads up and feedback period.

For that reason we also did not add any kickstart support for this override - dropping something from the GUI is of course not without impact, but doing the same for a kickstart command or option is much harder.

Also as already mentioned above, enabling password based root login over SSH can quite significantly compromise the security of a system and should be an explicit and easy to spot action performed by the user during the installation. Clicking a checkbox satisfies this condition in the GUI. A kickstart command option on the other hand could be easily missed  during the common practice of reusing kickstarts and kickstart pieces - there are already quite a few options even just for the `rootpw` command and one more option copy pasted from a test-run kickstart could easily by missed & turn all production image installs vulnerable to remote password guessing attack.

If you really need to enable password based SSH root login, you can just easily use the following two line %post script (one line without comments):

```
%post
# permit root login via SSH with password authetication
echo "PermitRootLogin yes" > /etc/ssh/sshd_config.d/01-permitrootlogin.conf
%end
```

This does 100% the same as a `rootpw` command option would, but unlike the option is quite explicit about what it does and easy to spot in a kickstart file.

Or even better, use the [sshkey command](https://pykickstart.readthedocs.io/en/latest/kickstart-docs.html#sshkey) to use a key instead of password.

To summarize a recommend not introducing the rootpw option and instead document the issue in the documentation, including the easy to use `%post` script based workaround.

bcl:
Thanks for the excellent summary @M4rtinK, I hadn't remembered that the checkbox was a workaround for that change. With that I'd have to completely agree and NAK this patch. 

@kkang-wr thanks very much for the PRs and the discussion, but as you can see it just isn't a safe thing to do, so if someone *really* wants to do it they need to make the extra effort.

kkang-wr:
@M4rtinK, thanks for explanation of the backgroud. I know it is easier to add an option than to remove it. If the checkbox will be removed from anaconda sooner or later, it is clearly improper to have this option in pykickstart. Thanks.

PR #376, opened 2021-03-15
Remove python six library

bcl:
This was necessary when transitioning from Python 2 to 3, but now that
pykickstart is Python 3 only it is just extra code that is unneeded.


poncovka (approved):
Looks good to me. Thanks!

VladimirSlavik (approved):

PR #378, opened 2021-03-28
[repo]Adding priority and modules_hotfixes params

claneys:
As cost is already part of the supported parameter, this commit just adds the support for priority which could be useful when using repositories that could provide the same packages. 

And module_hotfixes may be not as useful but it allows to supports hotfixes repositories.
Thoses repositories could be used in cases of creating images regurlarly in a CI system.


bcl:
Is there an associated Anaconda PR and discussion? That's where changes in behavior need to be implemented.
ISTR priority being discussed before and I think we decided that cost was good enough, or that it has no effect. I'll see if I can dig up the conversation.

If we decide to take these changes:

* Please don't rewrap lines.
* Please make separate commits for each change.

claneys:
Ok, no problem.

About priority, I'm agree with you it could be avoided but I thought it could be useful as well as the priority in dnf.conf files. In addition, it is clearer to use than cost but that's a bit subjective.

Ok, I'll unwrap the lines, this was made automatically by autopep8 with my editor. Is just trimming the trailing space is fine ?

pykickstart/commands/repo.py
```diff
@@ -488,6 +504,30 @@ class F33_Repo(F30_Repo):
...
+
+        return op
+
+class F35_Repo(F33_Repo):
+    removedKeywords = F33_Repo.removedKeywords
+    removedAttrs = F33_Repo.removedAttrs
+
+    def _getParser(self):
+        op = F33_Repo._getParser(self)
+        op.add_argument("--priority", type=int, version=F35, help="""
```
bcl, line 519: Wrong commit :)

claneys: my bad. Thanks for the review.

claneys:
> Is there an associated Anaconda PR and discussion? That's where changes in behavior need to be implemented.
> ISTR priority being discussed before and I think we decided that cost was good enough, or that it has no effect. I'll see if I can dig up the conversation.
> 
> If we decide to take these changes:
> 
> * Please don't rewrap lines.
> * Please make separate commits for each change.

The anaconda PR is on its way. See the mentioned PR above :)

poncovka:
Hi @claneys, could you try the example below and check if it fixes your issue? We get a lot of feature requests for the `repo` command, but we don't really want to support every repo configuration option. It would be better to find a more generic solution.

```
# Enable the custom repo.
repo --name "my-custom-repo"

%pre
# Generate the custom repo file.
cat >> /etc/anaconda.repos.d/custom.repo << EOF

[my-custom-repo]
name=My Custom Repository
baseurl=http://my/custom/repo/url/
priority=10
module_hotfixes=1

EOF
%end

```

VladimirSlavik:
This trick might warrant addition to anaconda docs, under common bugs or something similar... (common RFEs?)

claneys:
Ok, I'll check and let you know

claneys:
The trick is fine but do not fine our needs sadly. We are using our Koji infrastructure to build some images and such and then we use kickstart. But Koji get rid of the repo commands from the file and replace them by its own or from the ones from the command-line.

jkonecny12:
In that case I don't quite understand you change request. If Koji replaces the `repo` command it will replace also your new parameters for the `repo` KS command. Is that Koji logic even more complex?

claneys:
The idea after changing pykickstart and anaconda is also to change Koji. But those changes are quite heavier and not ready to be upstreamed. Koji do not handle priority at all at the moment, but we thought that those first pull request could be a start.

Conan-Kudo:
Having `priority=` and `module_hotfixes=` supported in pykickstart is absolutely a required starting point, otherwise everything else can't integrate support for it.

jkonecny12:
I would really like to find a better solution than extending API for `repo` command. The problem with repo command is that it's getting bigger and bigger and basically just slowly reflecting all repo file supported values. There is no reason to have all the API in pykickstart to just set configuration to DNF. The `repo` command should hold just the minimum required by Anaconda to set the installation running. In other words, these commands you are pointing here are great for people creating DVDs or have other special use case but not really that necessary for people doing installations on their machines.

I really hoped that generating repo file in the `%pre` section would cover all these scenarios. Unfortunately, seems that this does not work because of Koji. So maybe finding other solution to cover this?

From top of my mind, don't take me too seriously here, we may add a new pykickstart section something like `repo-config` which will take text and create the repo file and use that? We will not validate what you are writing there and it's your responsibility to make this working correctly.

Any better idea?

jkonecny12:
Also couldn't the `Koji` just remove the modify `repo` command feature it's doing now or make that optional?

claneys:
> Also couldn't the `Koji` just remove the modify `repo` command feature it's doing now or make that optional?

You could either choose to let it configure your repos or you can specify on the command line which repo you want, in any cases the repos are not taken from the ks file. The idea behind is that the repo are always dynamic because koji build continously packages and you can"t rely on 1 url or mirrorlist, you want to rely on working repositories that will be released once finished. Since this is the base of the release not base on an existing release.

jkonecny12:
Thanks for the explanation. I guess we have to find a better way in that case.

My question is, do you want to use your custom repository link or just be able to set the `priority` and `hotfixes` and use repository provided by Koji?

claneys:
Ok,
I hope we'll find out a solution. The goal is to be able to set priority and hotfixes yes. The repository link isn't a problem.

Conan-Kudo:
@jkonecny12 And I need this to land in order to wire in support for this for livecd-tools and appliance-tools, which people use to create media for users.

Conan-Kudo:
It's _probably_ possible, but currently `python-imgcreate` wires up pykickstart `repo` flags directly to DNF's API, so proper parameterization is preferred. If pykickstart itself turns it into a dictionary or whatnot internally, then it'd be easier to integrate such a thing.

jkonecny12:
We need to find a way how to solve this even in Anaconda. So you should be able to use a similar solution I guess.

Conan-Kudo:
Yeah, I suspect if it can be made to work for Anaconda without resorting to dumping files somewhere and pointing DNF to it, then it can work for imgcreate as well.

jkonecny12:
I will think about that and discuss that in a team to find the best approach to this.

Conan-Kudo:
Please make sure `gpgcheck` and `repo_gpgcheck` are also included in this.

claneys:
> I'm thinking about creating a new option (could be even for the `repo` command) which would enable you to add your settings in free form. Basically, we would then change this to what the DNF library understand. This would support everything what is supported by DNF `Repo` object (repo file). This option would be of course for advanced users only and wouldn't be checked for typos. It would be a bit similar to `authselect` command we have now.
> I have to look how hard would be to support something like that. But first, would that be feasible solution for you @claneys , @Conan-Kudo ?

Sorry for the delay, wasn't here for a while.
It would fit, but as you both pointed out my concerns are about the integration in anaconda to get them.

jkonecny12:
I asked DNF team about the possible solution and they told me that they have something like that already for parsing options from command line. However, it's not part of the API, so I'm going to file a bug on them and when they add this API it should be simple to implement something like `--dnf-option "priority=20" --dnf-option "module_hotfixes=True"` and we would then just forward that to DNF without checking if it is valid.

Conan-Kudo:
Call it `--repo-option` and I'm fine with that (let's not encode "dnf" into there when it's not necessary 😉 )

jkonecny12:
Bug on DNF: https://bugzilla.redhat.com/show_bug.cgi?id=1967925 .

jkonecny12:
PR to DNF with new API: https://github.com/rpm-software-management/dnf/pull/1779

bcl:
Closing this since we seem to have a consensus on adding a generic `--repo-option` option.

PR #379, opened 2021-04-19
Document missing feature '|' for ignoredisk and clearpart

jkonecny12:
It would be great to have this documented to avoid forgetting this again in the future. This functionality is handled by Anaconda.

This feature is documented in the Fedora documentation:

https://docs.fedoraproject.org/en-US/fedora/f33/install-guide/appendixes/Kickstart_Syntax_Reference/#sect-kickstart-commands-clearpart

We should definitely have it documented in the upstream too.


bcl (changes requested):
Take a look at the test failures.

pykickstart/commands/clearpart.py
```diff
@@ -151,7 +151,23 @@ def _getParser(self):
...
-class F28_ClearPart(F21_ClearPart):
+class F25_ClearPart(F21_ClearPart):
+    def _getParser(self):
+        op = super(F25_ClearPart, self)._getParser()
+        op.add_argument("--drives", type=commaSplit, version=F25, help="""
+                        You can combine multiple ways of specifying storage
+                        targets in one command, using the pipe ("|") as a
+                        delimiter. For example:
+
+                        ``clearpart --drives=sd*|hd*|vda,/dev/vdc``
```
poncovka, line 162: It is supported also for `--list`.

pykickstart/commands/ignoredisk.py
```diff
@@ -146,15 +146,35 @@ def _getParser(self):
...
     removedAttrs = F14_IgnoreDisk.removedAttrs
 
+    def _getParser(self):
+        op = F14_IgnoreDisk._getParser(self)
+        op.add_argument("--drives", type=commaSplit, version=F25, help="""
+                        You can combine multiple ways of specifying storage
+                        targets in one command, using the pipe ("|") as a
+                        delimiter. For example:
+
+                        ``ignoredisk --drives=sd*|hd*|vda,/dev/vdc``
```
poncovka, line 160: It is supported also for `--only-use`.

pykickstart/commands/clearpart.py
jkonecny12: Fixed.

pykickstart/commands/ignoredisk.py
jkonecny12: Fixed.

pykickstart/commands/clearpart.py
```diff
@@ -151,7 +151,23 @@ def _getParser(self):
...
+
+                        The above will match either of sd*, hd*, or vda and
+                        /dev/vdc. Variants are processed from left to right,
+                        and at least one of them must match for the command
+                        to succeed.""")
```
pcahyna, line 167: I must say, it is not at all clear what is the difference between `|` and `,`. It sounds like one is "or", while the other is "and", but the text says "vda and /dev/vdc", and no disk will be simultaneously named "vda" and "/dev/vdc", so it seems that the `vda,/dev/vdc` part will be simply ignored (will never match).

jkonecny12:
**UPDATED:**
* Fixed tests (great tests BTW)
* Fixed missing params which are also supported. Thanks @poncovka for pointing me on these!

pykickstart/commands/clearpart.py
jkonecny12: I had the same feeling so the message is definitely not great. We will try to improve the documentation the make the difference obvious.

In short, the difference is that `|` will match X-N disks for the whole expression. However, `,` must match all the time. In other words:

`sd*|vd*` could match - `vda`
`sda,vda`  have to match `sda` but also `vda`

VladimirSlavik: The option takes a list of drives. You specify that as a list of patterns, separated by comma `,` . Each pattern can consist of multiple alternatives, separated by pipe `|`. Patterns can also use asterisk `*` for wildcard matching.

As an example, `ignoredisk --drives=sd*|hd*|vda,/dev/vdc` has one comma, so it lists two drives: The first is either `sd*`, or `hd*`, or `vda`. The second is `/dev/vdc`.

jkonecny12:
Converting to draft to find out better explanation of the feature.

jkonecny12:
**UPDATED:**
* Now hopefully it's clear what it is doing.

I also wonder if we don't want to have a section above in the doc and just link to that section to avoid copy-paste issues. However, not sure if we want to have a section in the header just for four parameters in two commands.

poncovka (commented):
It looks great. Thanks!

bcl (changes requested):
This is better, and I don't mind the duplication since users tend to jump to the command they are interested in directly.

But I think it would be better to word it as an example, I am not sure I really understand what will happen after reading these. What is 'succeed' in each command's context?

It would be clearer to me if it said something like `clearpart --drives=/dev/vdc,sd*|hd*|vda` will clear partitions from vda, vdc, and all sd and hd drives in the system (if that's true). And tailor them to the command being documented.

jkonecny12:
So, do I understand it correctly that you want to remove text with explanation and just add an example instead. Because example is there even now.

> It would be clearer to me if it said something like clearpart --drives=/dev/vdc,sd*|hd*|vda will clear partitions from vda, vdc, and > all sd and hd drives in the system (if that's true). And tailor them to the command being documented.

By this example it's not clear what is the difference between `,` and `|` why not just having `/dev/vdc,sd*,hd*,vda`.

jkonecny12:
I've added another try to improve the situation. Do you like this version more or less @bcl ?

bcl:
When comparing these to the previous help strings I also see that we've totally lost what these options actually do and are focused on the format of the patterns. The previous help made sense to me, it was just missing the fact that a pattern could use wildcards and |

I think something more like this would be clearer (to me at least):


    Specifies which drives to clear partitions from. For
    example, the following clears the partitions on the
    first two drives on the primary IDE controller::
    ``clearpart --all --drives=sda,sdb``
    or to clear partitions on vda, and all sd and hd drives::
    ``clearpart --all --drives=vda,sd*|hd*``

Although now that I write that I realize I don't really understand why the pipe is used at all. Isn't this the same as `vda,sd*,hd*`?

Maybe we do need a separate section for this stuff and a pointer to it.

I think part of my problem understanding this is that when you say 'At least one match must be successful.' I am not sure of exactly what happens if it doesn't match anything. Similarly does 'Each pattern must match at least one drive, or it fails.' mean the kickstart will return an error, or that it just won't clear any drives?

I like the previous help strings because they told me what the action was for those examples. That seems to fit better with how I think :)

jkonecny12:
> When comparing these to the previous help strings I also see that we've totally lost what these options actually do and are focused on the format of the patterns. The previous help made sense to me, it was just missing the fact that a pattern could use wildcards and |
> 
> I think something more like this would be clearer (to me at least):
> 
> ```
> Specifies which drives to clear partitions from. For
> example, the following clears the partitions on the
> first two drives on the primary IDE controller::
> ``clearpart --all --drives=sda,sdb``
> or to clear partitions on vda, and all sd and hd drives::
> ``clearpart --all --drives=vda,sd*|hd*``
> ```
> 
> Although now that I write that I realize I don't really understand why the pipe is used at all. Isn't this the same as `vda,sd*,hd*`?

Yes, that is happening everyone and that is also reason why it's so hard to describe it. The difference is that Anaconda will fail to execute the command (installation will fail) when there is no `sdX` or `hdX` disk. So by doing `sd*|hd*` you are able to work on machines where is only `sdX` disks but the same kickstart can be used for machines with just `hdX` or both.

> 
> Maybe we do need a separate section for this stuff and a pointer to it.

I was thinking about that too. However, this rule applies only for the `ignoredisk` and `clearpart` so it could be confusing for people.


So should I use the old version or should I try to create a new other version?

bcl:
I think I have this figured out. I hope :) I think we need a paragraph at the top of clearpart and ignoredisk instead of repeating it for each option. I think it should say:

Optional commands that accept lists of drives or partitions can list specific device names, like sda, vda, sda1. Wildcards are supported using the ``*`` character. ``vda*`` will match all partitions on a virtio block device. Some of the commands, like ``--drives`` will fail if the device is not present. You can specify alternate names using the ``|`` pipe character. For example, to match sda or vdb you would use ``sda|vdb``, or to match any virtio block device but only sdb you would use ``vd*|sdb``.

Use the same paragraph for both commands.

And then in the --drives, --list, and only-use sections you don't need to repeat that but can now include a short example.

the following clears the partitions on the first two drives on the system: 
``clearpart --drives=sda,sdb``
or clear the first disk of systems using either scsi or virtio block devices:
``clearpart --drives=sda|vda``
clear all virtio block devices, or only the first scsi device:
``clearpart --drives=sda|vd*``

And just add some short examples to each of the others.

Does that make sense?

jkonecny12:
You are not 100% correct, gosh what have we created... :(

I agree with your suggestion. I've tried to implement that and hopefully explained it correctly now... Please look on just two last commits. I will squash everything when we get to the end but now, I would like to have history. What do you think?

docs/kickstart-docs.rst
```diff
@@ -130,6 +130,13 @@ You could use an entry similar to the following:
 
 ``ignoredisk --drives=/dev/disk/by-path/pci-0000:00:05.0-scsi-*``
 
+However, the ``*`` pattern will fail the installation in case no device could
```
bcl, line 133: Slight rewording:
However, the installation will fail if the ``*`` pattern does not match a device.

docs/kickstart-docs.rst
```diff
@@ -130,6 +130,13 @@ You could use an entry similar to the following:
...
+be matched. For ``clearpart`` and ``ignoredisk`` commands specifically you
+can also use ``|`` to solve this issue. For example to match ``vda`` and/or
+``hda`` you can use ``vda|hda`` which would match ``vda`` or ``hda`` or both.
+If nor ``vda`` or ``hda`` disks are not available the installation will fail.
```
bcl, line 137: Rewording:
But if neither ``vda`` or ``hda`` are available the installation will fail.
If you want to match any ``vd`` or ``hd`` disk you could combine ``*`` and ``|`` to match any drive of either type: ``vd*|hd*``

jkonecny12:
**UPDATED:**
* Implement suggestions from @bcl,  Thanks!
* Squashed the commits.

bcl:
Thanks for all the work!

PR #380, opened 2021-04-23
Update kickstart-docs.rst

tasleson:
Clarify that /dev/disk/by-path are not 100% reliable.  Add section
on selecting disks by size for installations which have consistent
disk configuration.

Signed-off-by: Tony Asleson <tasleson@redhat.com>

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -12,6 +12,7 @@ Contents:
    :maxdepth: 2
 
    Kickstart documentation <kickstart-docs>
+   Kickstart examples <kickstart-examples>
    Testing pykickstart <testing>
 
 Indices and tables
--- a/docs/kickstart-docs.rst
+++ b/docs/kickstart-docs.rst
@@ -105,7 +105,7 @@ device node name (such as ``sda``). The Linux kernel has moved to a more
 dynamic method where device names are not guaranteed to be consistent
 across reboots, so this can complicate usage in Kickstart scripts. To
 accommodate stable device naming, you can use any item from
-``/dev/disk`` in place of a device node name. For example, instead of:
+``/dev/disk/by-id`` in place of a device node name. For example, instead of:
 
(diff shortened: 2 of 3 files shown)


tasleson:
@dwlehman Please take a peek at this, thanks.

dwlehman:
The part about using `/dev/disk/by-id` is a substantial improvement over `by-path` (which is just as volatile as the node name).

The part about selecting disks by size might be better-suited for somewhere a bit less prominent. I can think of several other examples that are at least as compelling which, if they were all included in this doc, might substantially decrease readability. The maintainers will know better than I do whether the "by size" example is appropriate for this document or if there is a repository of examples that could house it.

bcl:
Nice! I like the idea of examples, but I think they belong  in a separate document. It's probably time to add one with various kickstart snippets that people can refer to. This keeps the main document focused on the most common uses.

VladimirSlavik:
@bcl Hi Brian, what do you expect from this? I'd be happy to help with this in some way, if I knew what to do. There doesn't seem to be a clear "do this to make it acceptable" or "let's not have it" message.

bcl:
> @bcl Hi Brian, what do you expect from this? I'd be happy to help with this in some way, if I knew what to do. There doesn't seem to be a clear "do this to make it acceptable" or "let's not have it" message.

I'd like to see a new document of kickstart snippet examples started so that we have a place to collect things like this. eg. Add a docs/kickstart-examples.rst and link to it from the docs/index.rst

tasleson:
> I'd like to see a new document of kickstart snippet examples started so that we have a place to collect things like this. eg. Add a docs/kickstart-examples.rst and link to it from the docs/index.rst

@bcl  Sorry I forgot about this one, suggested change added to PR and rebased

bcl:
Sorry to take so long, not sure why I didn't notice this was still open :/

PR #381, opened 2021-06-01
Add RHEL 9 handler

M4rtinK:
Add handler for RHEL 9, so that RHEL only commands and RHEL command
variations can be used in Anaconda on RHEL 9.

The RHEL 8 version of the handler was created by commits:
16ff122476ba01874a154b7e45204b8a71e6cdf5
b13b5a1e512c0705aa295c29526e8615da6e6296

Resolves: rhbz#1966730

### DIFF
--- a/pykickstart/commands/fcoe.py
+++ b/pykickstart/commands/fcoe.py
@@ -106,6 +106,9 @@ class F28_FcoeData(F13_FcoeData):
 class RHEL8_FcoeData(F28_FcoeData):
     pass
 
+class RHEL9_FcoeData(F28_FcoeData):
+    pass
+
 class F12_Fcoe(KickstartCommand):
     removedKeywords = KickstartCommand.removedKeywords
     removedAttrs = KickstartCommand.removedAttrs
@@ -191,4 +194,7 @@ class F28_Fcoe(F13_Fcoe):
         return op
 
 class RHEL8_Fcoe(F28_Fcoe):
+    pass
+
+class RHEL9_Fcoe(F28_Fcoe):
     pass
--- a/pykickstart/commands/autopart.py
+++ b/pykickstart/commands/autopart.py
@@ -649,3 +649,8 @@ class RHEL8_AutoPart(F29_AutoPart):
 
                     Partitioning scheme 'btrfs' was removed.""" % versionToLongString(RHEL8)
         return op
+
+
+class RHEL9_AutoPart(RHEL8_AutoPart):
+    pass
+
--- a/pykickstart/commands/bootloader.py
+++ b/pykickstart/commands/bootloader.py
@@ -513,3 +513,6 @@ class F34_Bootloader(F29_Bootloader):
         op = F29_Bootloader._getParser(self)
         op.remove_argument("--upgrade", version=F34)
         return op
+
+class RHEL9_Bootloader(F34_Bootloader):
+    pass
--- a/pykickstart/commands/btrfs.py
+++ b/pykickstart/commands/btrfs.py
@@ -276,3 +276,6 @@ class RHEL8_BTRFS(DeprecatedCommand, F23_BTRFS):
         op = F23_BTRFS._getParser(self)
         op.description += "\n\n.. deprecated:: %s" % versionToLongString(RHEL8)
         return op
+
+class RHEL9_BTRFS(RHEL8_BTRFS):
+    pass
--- a/pykickstart/commands/logvol.py
+++ b/pykickstart/commands/logvol.py
@@ -942,3 +942,6 @@ class RHEL8_LogVol(F29_LogVol):
 
                     Btrfs support was removed.""" % versionToLongString(RHEL8)
         return op
+
+class RHEL9_LogVol(RHEL8_LogVol):
+    pass
--- a/pykickstart/commands/ostreesetup.py
+++ b/pykickstart/commands/ostreesetup.py
@@ -90,3 +90,6 @@ class RHEL7_OSTreeSetup(F21_OSTreeSetup):
 
 class RHEL8_OSTreeSetup(F21_OSTreeSetup):
     pass
+
+class RHEL9_OSTreeSetup(F21_OSTreeSetup):
+    pass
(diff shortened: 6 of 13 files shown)


M4rtinK:
As @poncovka mentioned some BTRFS related adjustments will be needed, I am marking this as a draft for now.

Also we should definitely backport the final state of this PR to upstream branch to avoid issues in the next RHEL release.

VladimirSlavik (commented):
LGTM so far.

rvykydal:
> As @poncovka mentioned some BTRFS related adjustments will be needed, I am marking this as a draft for now.

We seem to need to create or use kickstart commands with support for btrfs removed. Relevant rhel-8 PRs:
https://github.com/pykickstart/pykickstart/pull/210
https://github.com/rhinstaller/anaconda/pull/2518
https://github.com/rhinstaller/anaconda/pull/1551

poncovka:
I think that we can just use the `RHEL8` versions to remove the Btrfs support:

```
RHEL8_AutoPart
RHEL8_BTRFS
RHEL8_LogVol
RHEL8_Raid
```

However, we will need a new `RHEL9_Partition` command that will be based on `F34_Partition`, but contain the logic of `RHEL8_Partition`.

These version will have to be changed also in Anaconda.

poncovka:
Based on the failing tests, I suggest to make some of the RHEL subclasses identical with their parents. Then, we don't have to create new commands every RHEL release:

```
RHEL8_Fcoe = F28_Fcoe
RHEL8_FcoeData = F28_FcoeData

RHEL7_OSTreeSetup = F21_OSTreeSetup
RHEL8_OSTreeSetup = F21_OSTreeSetup

RHEL7_VolGroup = F21_VolGroup
RHEL7_VolGroupData = F21_VolGroupData

RHEL8_VolGroup = F21_VolGroup
RHEL8_VolGroupData = F21_VolGroupData
```

Conan-Kudo:
Would it be possible to make it relatively easy to restore support for Btrfs? In the CentOS Hyperscale SIG, we're layering on CentOS Stream and enabling Btrfs so we can deliver a user experience leveraging it.

Conan-Kudo:
cc: @davide125, @michel-slm

bcl:
@M4rtinK 
> we go back to the old way of empty sub-classes for RHEL 9

That's the best way to go. By assigning the class to a new name you are only creating a new variable, not a new class and that's going to mess with pykickstart's internals which use the `__class__.__name__` for various things because both of them look like the same class.


@Conan-Kudo 
> Would it be possible to make it relatively easy to restore support for Btrfs?

Please open an issue about this, with details on what you plan to do to. I'm not how much work would be involved to support another namespace, we've never needed to do that with CentOS before -- but now with other RHEL based distributions appearing (Rocky Linux, AlmaLinux) it's probably time to think about it.

poncovka:
@bcl 

> That's the best way to go. By assigning the class to a new name you are only creating a new variable, not a new class and that's going to mess with pykickstart's internals which use the `__class__.__name__` for various things because both of them look like the same class.

In that case, couldn't we just remove these classes from pykickstart? This affects us. We have to change versions of these commands in Anaconda for no good reason: https://github.com/rhinstaller/anaconda/commit/d8fef2e7bde8860224bbf61512a6b57c5e1dfb9e

If others are not sure, what we're talking about, here is an example of the classes:

```
class RHEL7_VolGroup(F21_VolGroup):
    pass

class RHEL8_VolGroup(F21_VolGroup):
    pass
```

bcl:
> In that case, couldn't we just remove these classes from pykickstart?

No, they are needed to differentiate between RHEL and Fedora behavior. Right now they are empty because things haven't diverged much but when they do they become more important.

poncovka:
> No, they are needed to differentiate between RHEL and Fedora behavior. Right now they are empty because things haven't diverged much but when they do they become more important.

Therefore, they are not needed at all right now and could be created, when things will diverge. That is what we do with Fedora classes that we have in RHEL handlers. These RHEL commands might never diverge, but we will have to create them again and again.

In Anaconda, we are trying to keep the RHEL-ification of the code as small and simple as possible, so keeping these classes just in case, doesn't make sense to me. Anyway, we don't need to decide that now or change that in RHEL 9.

bcl (changes requested):
The RHEL9 handler should be added upstream first, then ported to the rhel9-branch (which hasn't diverged much so that should be easy). If anaconda wants to use empty classes or not that's fine but we should not make drastic changes when branching for a new release so they should be added at some point.

M4rtinK:
> The RHEL9 handler should be added upstream first, then ported to the rhel9-branch (which hasn't diverged much so that should be easy). If anaconda wants to use empty classes or not that's fine but we should not make drastic changes when branching for a new release so they should be added at some point.

Sounds good, to summarize:
- let's rebase this PR against the upstream master branch
- use the classic class inheritance for now
- reuse the BTRFS PRs from RHEL 8 for now
- after being merged upstream, rebase for RHEL 9

That should get what's needed short term in place & further improvemenets can be done by separate PRs later.

M4rtinK:
Changed base branch to `master` and updated the PR:
- rebased on `master` branch
- split to three commits
  * BTRFS related stuff
  * general RHEL command class version bumps
  * RHEL 9 handler

For the BTRFS stuff I've just sub classed the RHEL 8 command versions - I hope that's correct ? We did deprecate it in RHEL 8 & it's still deprecated. 

Only in the case of the `partition` command I had to duplicate the deprecation code as we did update the command in F34.

bcl (changes requested):
Could you add a test for the new RHEL9 partition class? Gotta keep those coverage numbers moving in the right direction ;)

pykickstart/commands/fcoe.py
```diff
@@ -106,6 +106,9 @@ def _getArgsAsStr(self):
 class RHEL8_FcoeData(F28_FcoeData):
     pass
 
+class RHEL9_FcoeData(RHEL8_FcoeData):
```
rvykydal, line 109: Shouldn't  we inherit from Fedora class (F28_FcoeData) here ? At least to be consistent with other commands created in this patch.

M4rtinK: Good point, fixed!

M4rtinK:
> Could you add a test for the new RHEL9 partition class? Gotta keep those coverage numbers moving in the right direction ;)

Good point, did that and looks like the coverage test is happy  now. :)

rvykydal (approved):
Looks good to me

bcl (approved):
Looks good, thanks!

Issue #382, opened 2021-06-04
RFE: ksvalidator: Support multiple files and printing filename


ggiesen:
It would be great if ksvalidator supported passing multiple filenames and/or a shell glob, as well as an option to print the filename in order to be able to use it as a linting tool in a CI/CD pipeline

VladimirSlavik:
This is what I get with code in #383 - does that work for you @ggiesen?

```
$ PYTHONPATH=.:$PYTHONPATH ./tools/ksvalidator.py http://cobra02/u/vs/ks/default.ks http://cobra02/u/vs/ks/anaconda-ks.cfg hello*

Checking kickstart file http://cobra02/u/vs/ks/default.ks


Checking kickstart file http://cobra02/u/vs/ks/anaconda-ks.cfg

The option --isUtc will be deprecated in future releases. Please modify your kickstart file to replace this option with its preferred alias --utc.

Checking kickstart file helloblah.ks

The following problem occurred on line 1 of the kickstart file:

Unknown command: foo


Checking kickstart file helloworld.ks

The option --isUtc will be deprecated in future releases. Please modify your kickstart file to replace this option with its preferred alias --utc.
```

ggiesen:
That's perfect.

VladimirSlavik:
#383 was merged, so I think this could be closed?

PR #383, opened 2021-06-08
ksvalidator - multiple files

VladimirSlavik:
A try at #382 - multiple files and globbing. Thus far, it's a quick hack to see if it's even possible.

Unfortunately there's something mysterious about the way warnings are handled, I can't divert them no matter what. Neither wrapping in `catch_warnings` nor replacing `showwarning` does anything.

This is only a draft, so no tests.

### DIFF
--- a/tools/ksvalidator.py
+++ b/tools/ksvalidator.py
@@ -48,8 +48,8 @@ def cleanup(dest, fn=None, exitval=1):
     return exitval
 
 def main(argv):
-    op = argparse.ArgumentParser(usage="%(prog)s [options] ksfile", add_help=False)
-    op.add_argument("ksfile", nargs="?",
+    op = argparse.ArgumentParser(usage="%(prog)s [options] ksfile [ksfile...]", add_help=False)
+    op.add_argument("ksfile", nargs="*",
                     help=_("filename or URL to read from"))
     op.add_argument("-e", "--firsterror", dest="firsterror", action="store_true",
                     default=False, help=_("halt after the first error or warning"))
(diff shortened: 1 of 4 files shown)


bcl:
Do you mean the 'Found option with uppercase letters' from tests/args_names.py? Those are prints not something that gets raised. They aren't a problem as long as they have a lower case version, in which case the test will fail.

tools/ksvalidator.py
```diff
@@ -80,42 +81,62 @@ def main(argv):
...
+                retval += 1
+                retmsg.append(_("General kickstart error in input file"))
+            except Exception as e:        # pylint: disable=broad-except
+                retval += 1
+                retmsg.append(_("General error in input file:  %s") % e)
+
+    return retval, retmsg
 
 if __name__ == "__main__":
     retval, messages = main(sys.argv[1:])
```
bcl, line 142: Pylint doesn't like retval here and in main, rename it to 'rc' or similar.

tools/ksvalidator.py
```diff
@@ -80,42 +81,62 @@ def main(argv):
...
+            ksfiles.extend(glob.glob(inksfile))
+
+    # check if there are any files to check
+    if not ksfiles:
+        return (1, "No files match the patterns.")
+
+    # iterate over files to check them
+    with tempfile.TemporaryDirectory(prefix="ksvalidator-tmp-") as destdir:
+        for ksfile in ksfiles:
+            print(_("Checking kickstart file %(filename)s\n" % {"filename": ksfile}))
```
bcl, line 102: I am not sure why pylint is complaining about this %, we have _() strings with % in them all over the place.

bcl: Hm. pocketlint should be keeping that warning from showing up, maybe something odd with the GH workflow setup.

tools/ksvalidator.py
VladimirSlavik: Done.

tools/ksvalidator.py
VladimirSlavik: Doesn't show with last changes, probably a fluke of some kind.

bcl:
Looks good to me, just needs some more tests to keep that coverage from dropping ;)

VladimirSlavik:
I don't think I understand the coverage thing. On the site it shows everything stayed the same except for the one file which went up, but here I still see it go down.

bcl:
> I don't think I understand the coverage thing. On the site it shows everything stayed the same except for the one file which went up, but here I still see it go down.

Yeah, that's odd. It appears to be the difference between branches covered. On the last master run it says " 2248 of 2390 branches covered (94.06%) " and on this PR it is saying " 2255 of 2398 branches covered (94.04%) " but I can't see any differences in the report so I think we can ignore that small drop :)

bcl (approved):
Looks good to me, let me know if you're ready to have this merged or want to wait for feedback from #382

VladimirSlavik:
Thanks! Not yet, I think.
- The docs/manpages need updating too.
- I just realized there should be a check if the globs actually match anything.

VladimirSlavik:
This is not dead! Rebased on master to start with a resuscitation and resolve conflict with #391.

VladimirSlavik:
Docs are now fine, and I can't really understand the other point about no files found from globs, so I think this is fine?

bcl (approved):
Looks good, thanks!

PR #387, opened 2021-07-09
remove authconfig support

pbrezina:
Authconfig compatibility tool (from authselect-compat) will be removed from Fedora 35:
https://fedoraproject.org/wiki/Changes/RemoveAuthselectCompatPackage

### DIFF
--- a/pykickstart/commands/authconfig.py
+++ b/pykickstart/commands/authconfig.py
@@ -15,14 +15,14 @@
 # Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.  Any Red Hat
 # trademarks that are incorporated in the source code or documentation are not
 # subject to the GNU General Public License and may only be used or replicated
-# with the express permission of Red Hat, Inc. 
+# with the express permission of Red Hat, Inc.
 #
 import warnings
 from textwrap import dedent
 
 from pykickstart.errors import KickstartDeprecationWarning
-from pykickstart.version import FC3, versionToLongString, F28
-from pykickstart.base import KickstartCommand
+from pykickstart.version import FC3, versionToLongString, F28, F35
+from pykickstart.base import KickstartCommand, RemovedCommand
 from pykickstart.options import KSOptionParser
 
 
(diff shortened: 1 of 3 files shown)


bcl:
This isn't the correct way to remove a command. We have a class for that :) See commit ab95215705e318e6a8e5d380e09da976e13b5f29 for an example.

pbrezina:
Thanks for the hint. See new patch.

pykickstart/commands/authconfig.py
```diff
@@ -22,7 +22,7 @@
 
 from pykickstart.errors import KickstartDeprecationWarning
 from pykickstart.version import FC3, versionToLongString, F28
```
bcl, line 24: You need to import F35.

Also, you can run `make check` and `make test` locally to check the tests before pushing.

bcl (changes requested):

pykickstart/commands/authconfig.py
pbrezina: This should be fixed now. Make check yields many errors like:
```
************* Module /home/pbrezina/workspace/pykickstart/tests/commands/network.py
E1102(not-callable):116,14: FC3_TestCase.runTest: self.handler is not callable
```

Maybe I run it wrong?

VladimirSlavik (approved):
Not sure about the coverage, but looks good to me.

pykickstart/commands/authconfig.py
bcl: Possibly a different pylint version? I'm running on Fedora 33 with python3-pylint-2.6.0-2.fc33.noarch and don't see it complain about that. Running with tox also doesn't return that error so I'm not sure what's going on with your setup.

bcl (approved):
Looks good, thanks!

VladimirSlavik:
@pbrezina, I'm curious, why did you make the initial patch just a plain removal of stuff from the handler?

The reason to ask is, I re-checked the docs for removing a command and noticed that's mentioned there as the first thing. The "correct new way" is mentioned as an alternative. So if that text was the reason for how this looked, then amending the doc would be a good idea.

pbrezina:
Honestly, I was just too lazy and did not check any docs.

VladimirSlavik:
Fair enough!

PR #390, opened 2021-09-27
Fix newer pylint complaints

bcl:


VladimirSlavik (approved):
LGTM.

Personally I'd say the u-strings are wrong, but it's easy to understand why not touching them is the better idea...

bcl:
> Personally I'd say the u-strings are wrong, but it's easy to understand why not touching them is the better idea...

Right, there's enough of them to make it annoying to change, and really they aren't hurting anything at the moment.

PR #391, opened 2021-09-27
ksvalidator: Handle empty files

### LINKED ISSUE #388: ksvalidator: weird error on empty file
Problem:

When validating empty file, user gets a very confusing and un-helpful error:

    General error in input file:  stat: path should be string, bytes, os.PathLike or integer, not NoneType

To reproduce:

    $ touch foo.ks
    $ ksvalidator foo.ks
    General error in input file:  stat: path should be string, bytes, os.PathLike or integer, not NoneType
    $ 

Expectation:

User should get a meaningful message, eg. about missing mandatory content.
Or no error (see note below).

Version:

    $ rpm -q pykickstart
    pykickstart-3.32-2.fc34.noarch
    $ 

Note that if the file contains eg. single space, ksvalidator does not complain and
considers file OK:

    $ echo -n ' ' > foo.ks
    $ ksvalidator foo.ks; echo $?
    0
    $

bcl:
Previously an empty file would return a confusing error. This fixes it
by checking the preprocess call for None, and by running strip on the
file before checking if it is empty. Also includes tests.

Fixes #388

### DIFF
--- a/pykickstart/parser.py
+++ b/pykickstart/parser.py
@@ -124,7 +124,7 @@ def preprocessFromString(s):
        run.  Returns the location of the complete kickstart file.
     """
     s = preprocessFromStringToString(s)
-    if s:
+    if s.strip():
         import tempfile
         (outF, outName) = tempfile.mkstemp(suffix="-ks.cfg")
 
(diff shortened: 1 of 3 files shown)


bcl:
Ignore the covscan failure. Coverage can't cover __main__ code.

VladimirSlavik (approved):
Looks good to me. Thank you!

PR #394, opened 2021-11-24
Support for RHEL9 "inst.repo" instead of older "repo"

tkopecek:
There are more inst.* commands which are now required (f34+?), so maybe it makes more sense to create F34 or F35 classes? Not sure, this is just a simple patch for koji needs but it probably should be elaborated more.


VladimirSlavik:
Thanks for the attempt. However, `inst.repo` is a boot option, this is kickstart. Apples and oranges. Or is there something I am missing?

tkopecek:
OMG, sorry for noise. It fixed some things for me but it was obviously just masking underlying issue with oz/imagefactory sending old parameters to kernel as you've pointed out. Dropping.

lhh:
Excellent, thanks!

PR #406, opened 2022-04-06
i18n: Pass gettext domain on every translation call

### LINKED ISSUE #405: Don't set the global gettext domain
Pykickstart shoudn't call `gettext.textdomain("pykickstart")` and rely on using the global domain. If the global domain is changed, the pykickstart translations will no longer work. The domain should be specified for every gettext call.

For example:
https://github.com/rhinstaller/anaconda/blob/master/pyanaconda/core/i18n.py#L25
https://github.com/storaged-project/blivet/blob/3.5-devel/blivet/i18n.py#L40

bcl:
Previously it was changing the global gettext domain which could cause
problems for other applications. This updates it to pass the domain to
every call, and also switches to using the more robust .translation()
method as used by Anaconda.

Fixes #405

### DIFF
--- a/pykickstart/i18n.py
+++ b/pykickstart/i18n.py
@@ -25,7 +25,6 @@ def _find_locale_files():
     locale_path = os.path.join(os.path.dirname(module_path), 'locale')
 
     gettext.bindtextdomain("pykickstart", locale_path)
-    gettext.textdomain("pykickstart")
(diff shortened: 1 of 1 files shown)


poncovka (approved):
Thank you!

VladimirSlavik (approved):
Thank you!

PR #412, opened 2022-06-15
Write commands in alphabetical order

bcl:
When a kickstart is output the commands are grouped by write priority,
then alphabetical within the group. This isn't how it was actually
working -- it was using the full class name with release prefixes (eg.
F37_) so a newer command would sort before older commands, as seen with
the recent rootpw change.

This changes the insert to compare based on the command name portion of
the class name, without the prefix. This will change the order of some
things -- but in the future the output will be in stable alphabetical
order.

Resolves: rhbz#2096871

### DIFF
--- a/pykickstart/base.py
+++ b/pykickstart/base.py
@@ -305,6 +305,12 @@ class KickstartHandler(KickstartObject):
         return retval
 
     def _insertSorted(self, lst, obj):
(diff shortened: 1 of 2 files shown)


poncovka (approved):
Thank you!

VladimirSlavik (approved):
Looks good to me. Thank you!

PR #420, opened 2022-08-19
Add --hibernation option to AutoPart

OndrejZobal:
Add F38_AutoPart class with a new Kickstart flag that creates a swap partition for hibernation.

Installer implementation: https://github.com/rhinstaller/anaconda/pull/4275

### DIFF
--- a/pykickstart/commands/autopart.py
+++ b/pykickstart/commands/autopart.py
@@ -19,7 +19,7 @@
 #
 from pykickstart.base import KickstartCommand
 from pykickstart.version import versionToLongString, RHEL6, RHEL7, RHEL8
-from pykickstart.version import FC3, F9, F12, F16, F17, F18, F20, F21, F26, F29
+from pykickstart.version import FC3, F9, F12, F16, F17, F18, F20, F21, F26, F29, F38
 from pykickstart.constants import AUTOPART_TYPE_BTRFS, AUTOPART_TYPE_LVM, AUTOPART_TYPE_LVM_THINP, AUTOPART_TYPE_PLAIN
 from pykickstart.errors import KickstartParseError
 from pykickstart.options import KSOptionParser
@@ -526,7 +526,8 @@ class F26_AutoPart(F23_AutoPart):
                         Do not create a /boot partition.""")
         op.add_argument("--noswap", action="store_true", default=False,
                         version=F26, help="""
-                        Do not create a swap partition.""")
+                        Do not create a swap partition. Only one of ``--noswap``
+                        and ``--hibernation`` can be specified.""")
         return op
 
 class F29_AutoPart(F26_AutoPart):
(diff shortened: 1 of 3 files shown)


bcl (approved):
Looks good to me.

VladimirSlavik (approved):
Looks good to me. Thank you!

pykickstart/commands/autopart.py
```diff
@@ -654,3 +654,38 @@ def _getParser(self):
 class RHEL9_AutoPart(RHEL8_AutoPart):
     pass
 
+
+class F38_AutoPart(F29_AutoPart):
+    def __init__(self, writePriority=100, *args, **kwargs):
```
poncovka, line 659: You don't have to redefine the write priority, just use `args` and `kwargs`.

pykickstart/commands/autopart.py
```diff
@@ -654,3 +654,38 @@ def _getParser(self):
...
+        F29_AutoPart.__init__(self, writePriority=writePriority, *args, **kwargs)
```
poncovka, line 660: You can call `super()` instead of `F29_AutoPart`. Same below.

pykickstart/commands/autopart.py
```diff
@@ -654,3 +654,38 @@ def _getParser(self):
...
+        self.hibernation = kwargs.get("hibernation", False)
+        self.swap = kwargs.get("swap", False)
```
poncovka, line 662: This line looks like a leftover.

pykickstart/commands/autopart.py
```diff
@@ -654,3 +654,38 @@ def _getParser(self):
...
+
+    def __str__(self):
+        retval = F29_AutoPart.__str__(self)
+        if not self.autopart:
+            return retval
+
+        if self.hibernation:
```
poncovka, line 669: You can check `if self.autopart and self.hibernation` and remove the condition above.

pykickstart/commands/autopart.py
```diff
@@ -654,3 +654,38 @@ def _getParser(self):
...
+            retval += " --hibernation"
+            retval += "\n"
+
+        return retval
+
+    def _getParser(self):
+        op = F29_AutoPart._getParser(self)
+        op.add_argument("--hibernation", action="store_true", default=False,
+                        version=F38, help="""
+                        Create larger swap for hibernation.""")
```
poncovka, line 681: Maybe it could say something like: `Create a swap partition with an automatically determined size that's big enough for hibernation.` You might also mention, that the `--hibernation` option shouldn't be used together with `--noswap`.

OndrejZobal: >  You might also mention, that the --hibernation option shouldn't be used together with --noswap.
Isn't this obvious to the average Kickstart user? Also If I include this warning in `--hibernation` I should probably add it to `--noswap` as well.

VladimirSlavik: Actually, yes, both places should have the note.

pykickstart/commands/autopart.py
```diff
@@ -654,3 +654,34 @@ def _getParser(self):
...
+            retval += " --hibernation"
+            retval += "\n"
+        return retval
+
+    def _getParser(self):
+        op = super()._getParser()
+        op.add_argument("--hibernation", action="store_true", default=False,
+                        version=F38, help="""
+                        Create a swap partition with an automatically determined
+                        size that's big enough for hibernation. Incompatable with --noswap.""")
```
poncovka, line 677: There is a typo.

poncovka (approved):
It looks good to me. Thank you!

poncovka (approved):

pykickstart/commands/autopart.py
```diff
@@ -526,7 +526,7 @@ def _getParser(self):
                         Do not create a /boot partition.""")
         op.add_argument("--noswap", action="store_true", default=False,
                         version=F26, help="""
-                        Do not create a swap partition.""")
+                        Do not create a swap partition. Incompatible with --hibernation.""")
```
bcl, line 529: Looking at other examples of exclusive options it looks like we have historically used the same wording as the runtime error, added to the help. So this should be similar to autopart's --pbkdf-time help string -- `Only one of ``--noswap`` and ``--hibernation`` can be specified.` Note the backtick usage here, they are used by sphinx when generating the docs.

bcl (changes requested):
Thanks, just one more change :)

bcl (approved):
Looks good, thanks!

PR #422, opened 2022-08-26
sshkey: Escapes quotes in the ssh key

bcl:
This allows spaces and quotes to be used in the options section of the
ssh key, eg:

    sshkey --username=root "from=\"192.168.1.1\" ssh-rsa ..."
    sshkey --username=user "\"/usr/bin/rrsync /home/user/public_html/\",no-port-forwarding ssh-rsa ..."

Resolves: rhbz#2117734

### DIFF
--- a/pykickstart/commands/sshkey.py
+++ b/pykickstart/commands/sshkey.py
@@ -58,6 +58,17 @@ class F22_SshKeyData(BaseData):
         retval += ' "%s"' % self.key
         return retval
 
+class F38_SshKeyData(F22_SshKeyData):
+    removedKeywords = BaseData.removedKeywords
+    removedAttrs = BaseData.removedAttrs
+
+    def _getArgsAsStr(self):
+        retval = ""
+
+        retval += " --username=%s" % self.username
+        retval += ' "%s"' % self.key.replace('"', r'\"')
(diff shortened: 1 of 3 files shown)


pykickstart/commands/sshkey.py
```diff
@@ -58,6 +58,17 @@ def _getArgsAsStr(self):
...
 
+class F38_SshKeyData(F22_SshKeyData):
+    removedKeywords = BaseData.removedKeywords
+    removedAttrs = BaseData.removedAttrs
+
+    def _getArgsAsStr(self):
+        retval = ""
+
+        retval += " --username=%s" % self.username
+        retval += ' "%s"' % self.key.replace('"', r'\"')
```
VladimirSlavik, line 69: I wonder if it would make sense to use `shlex.quote` - at a glance, that's the exact situation it was made for?

VladimirSlavik (approved):
Looks good to me. Thank you!

pykickstart/commands/sshkey.py
bcl: You would think so, wouldn't you :) But that's not actually what it does. shlex can unescape `\"` but it doesn't generate them.

PR #425, opened 2022-09-05
Update documentation for the rhsm command

M4rtinK:
Since RHEL 9 the --server-hostname is used to set the Satellite server
URL when registering to Satellite. Also clarify that when registering to
Satellite the --rhsm-baseurl is set automatically.


pykickstart/commands/rhsm.py
```diff
@@ -96,9 +96,9 @@ def _getParser(self):
                         ``--proxy=[protocol://][username[:password]@]host[:port]``
                         """)
         op.add_argument("--server-hostname", metavar="<subscription_service_server_url>", version=RHEL8, required=False,
-                        help="Red Hat subscription service server URL.")
+                        help="Red Hat Satellite server URL (registration to Satellite is supported since RHEL 9.0).")
```
ekohl, line 99: The parameter says hostname, metavar says URL and the help too. So what is the expected value? `satellite.example.com` or `https://satellite.example.com/`?

pykickstart/commands/rhsm.py
```diff
@@ -96,9 +96,9 @@ def _getParser(self):
...
         op.add_argument("--rhsm-baseurl", metavar="<content_base_url>", version=RHEL8, required=False,
-                        help="Content base URL.")
+                        help="Content base URL. Set automatically when registering to Red Hat Satellite.")
```
ekohl, line 101: So this argument really is the content base url? With RHSM I'd expect it to be Candlepin, not Pulp.

When `--server-hostname` is given, is the argument still respected or actually ignored?

bcl:
It might be better to just add a section to the command documentation mentioning the RHEL 9 change instead of trying to get more detailed in the argument docs.

ekohl:
I think @M4rtinK said that in RHEL 8 it was just for debugging purposes. If so, would it be correct to say that in RHEL 8 it did nothing for the user? If so, you could also say `VERSION=RHEL9`, possibly with a note that in RHEL 8 it could be used, but didn't affect anything.

M4rtinK:
> I think @M4rtinK said that in RHEL 8 it was just for debugging purposes. If so, would it be correct to say that in RHEL 8 it did nothing for the user? If so, you could also say `VERSION=RHEL9`, possibly with a note that in RHEL 8 it could be used, but didn't affect anything.

To clarify - RHEL 9+ is necessary if you want to register to Satellite via Anaconda instead of to Hosted Candlepin. Also since RHEL 9.0 the `--server-hostname` option of the `rhsm` is used to set the Satellite URL.

In comparison, RHEL 8 Anaconda **does not support registering to Satellite** - it won't fetch the provisioning script or do any of the other minor, yet vital things necessary for registration against Satellite instead of to Hosted Candlepin. It has the `--server-hostname` option but as mentioned, its only really useful for debugging during development against staging infrastructure.

M4rtinK:
> It might be better to just add a section to the command documentation mentioning the RHEL 9 change instead of trying to get more detailed in the argument docs.

Sounds good - I'll update the PR soon (tm) to do it like this. :)

pykickstart/commands/rhsm.py
M4rtinK: In RHEL 8, `--server-hostname` just sets the `hostname` key in the server section of the `rhsm.conf` via DBus. The `--rhsm-baseurl` just sets the `baseurl` key in the `rhsm` section of the config.

In RHEL 9 we still do the same - if --server-hostname or --rhsm-baseurl are present, we set the respective keys in the config file. But also do one extra thing - if --server-hostname is set we consider it to be a request to register to Satellite on the given URL: https://github.com/rhinstaller/anaconda/blob/rhel-9/pyanaconda/modules/subscription/runtime.py#L1025

So we do fetch the Satellite provisioning script and execute it + some other minor things.

The Satellite provisioning script usually sets the `baseurl` key in the `rhsm` section of `rhsm.conf` when executed, then indeed, the value provided to `--rhsm-baseurl` can be most likely considered as effectively ignored. 

This is because IIRC the values are first set in the RHSM config & then the script is executed, overwriting the baseurl:
https://github.com/rhinstaller/anaconda/blob/rhel-9/pyanaconda/ui/lib/subscription.py#L236
https://github.com/rhinstaller/anaconda/blob/rhel-9/pyanaconda/ui/lib/subscription.py#L279

ekohl: That helps a lot. Thanks!

PR #426, opened 2022-09-15
Fix syntax of a code sample

ekohl:
With reStructuredText the :: after example already start a code block.  This makes the backticks redundant and are shown in the rendered documentation.

https://pykickstart.readthedocs.io/en/latest/kickstart-docs.html#timezone
Screen Shot 2022-09-15 at 12 11 53


VladimirSlavik (approved):
Looks good to me. Thank you!

PR #427, opened 2022-10-28
bootloader: Add systemd-boot support with --sdboot

jlinton:
Extend the bootloader command with the ability to
utilize systemd-boot as the bootloader on EFI systems.

Signed-off-by: Jeremy Linton <jeremy.linton@arm.com>


bcl:
kickstart commands need a corresponding change in Anaconda, and Anaconda doesn't support systemd-boot, and as far as I know don't have any plans to add it.

But if this were to be added it would also need corresponding tests.

jlinton:
I am the plans :) For what that is worth, Lol.

The anaconda PR is mentioned above. There is a matching package (to own the config files and some cleanup), and a few bits scattered here/there that are dependent on the anaconda change or the package. Its a stack of changes.

jkonecny12:
Hi @bcl this is proposed to be a pure community feature so Anaconda team won't support this directly. I just want to give you a heads-up about this: 
https://github.com/rhinstaller/anaconda/pull/4368#issuecomment-1300575714

jlinton:
The test cases should now be updated to support the bootloader --sdboot option.

pykickstart/commands/bootloader.py
```diff
@@ -514,3 +514,28 @@ def _getParser(self):
...
+class F38_Bootloader(F34_Bootloader):
+    removedKeywords = F34_Bootloader.removedKeywords
+    removedAttrs = F34_Bootloader.removedAttrs
+
+    def __init__(self, writePriority=10, *args, **kwargs):
+        F34_Bootloader.__init__(self, writePriority, *args, **kwargs)
+        self.sdboot = kwargs.get("sdboot", False)
+
+    def _getArgsAsStr(self):
+        ret = F21_Bootloader._getArgsAsStr(self)
```
bcl, line 527: In `F21_Bootloader._getArgsAsStr` it returns only `--disabled` if it is disabled so I think we'll need to do that in all subsequent `_getArgsAsStr` functions as well. That behavior should also be tested in the new F38 test.

bcl (changes requested):
Looks pretty good, except for `--disabled` handling.

bcl:
If you rebase this on master those tests should pass.

pykickstart/commands/bootloader.py
jlinton: Ah, good catch. That is a bug for sure!

To avoid this problem in the future, and a lot of DRY, i stretched my python abilities and tweaked future class/method overriding.  :)

But the python docs say that @final is only in python 3.8, and I'm not sure if that is an appropriate assumption here?

jlinton: Yah, so I'm blowing up the unit test with the final... Sorta expected that to happen..

bcl:
Yeah, I think I'd rather just carry the disable check forward. Less surprising that way.

jlinton:
I reverted all the "fancy" stuff.  All future bootloader subclasses will just have to copy and paste the disable bit.

bcl (approved):
Looks good. I'll merge when Anaconda is ready for it.

poncovka:
Btw. I think that the combination of `--extlinux` and `--sdboot` options shouldn't be allowed.

poncovka:
The pull request is updated for Fedora 39, since we missed the beta freeze of Fedora 38. It can be merged and released.

PR #430, opened 2022-11-30
Update %post example for DNS problems

### LINKED ISSUE #429: Copying /etc/resolv.conf no longer works in %post
Thanks to systemd's takeover of /etc/resolv.conf you can no longer just copy the host/mock file into the chroot and have it work. Need to figure out what does work in order to enable DNS for those cases where it is needed in `%post` and update the documentation.

bcl:
Use the example from Anaconda's docs, and point to them.

Fixes #429


poncovka (approved):
Thank you!

PR #431, opened 2023-01-13
Add DNS handling options to the network command

VladimirSlavik:
This adds four new options to control handling of DNS:

--ipv4-dns-search
--ipv6-dns-search
--ipv4-ignore-auto-dns
--ipv6-ignore-auto-dns

Related: [rhbz#1656662](https://bugzilla.redhat.com/show_bug.cgi?id=1656662#c9)


VladimirSlavik:
Curious, the tests run ok locally on 3.10.8 from Fedora

bcl:
Whoops, didn't realize it would close it :) Rebase this on master and they should pass.

VladimirSlavik:
Thanks! Looks like it runs on a merge commit anyway, so all is well.

pykickstart/commands/network.py
```diff
@@ -280,6 +280,29 @@ def _getArgsAsStr(self):
...
 
+class F38_NetworkData(F27_NetworkData):
+    removedKeywords = F27_NetworkData.removedKeywords
+    removedAttrs = F27_NetworkData.removedAttrs
+
+    def __init__(self, *args, **kwargs):
+        F27_NetworkData.__init__(self, *args, **kwargs)
+        self.ipv4_dns_search = kwargs.get("ipv4_dns_search", None)
+        self.ipv6_dns_search = kwargs.get("ipv6_dns_search", None)
+        self.ipv4_ignore_auto_dns = kwargs.get("ipv4_ignore_auto_dns", None)
```
bcl, line 291: bools should default to False, not None.

pykickstart/commands/network.py
```diff
@@ -280,6 +280,29 @@ def _getArgsAsStr(self):
...
+        self.ipv6_ignore_auto_dns = kwargs.get("ipv6_ignore_auto_dns", None)
+
+    def _getArgsAsStr(self):
+        retval = F27_NetworkData._getArgsAsStr(self)
+        if self.ipv4_dns_search:
+            retval += " --ipv4-dns-search=%s" % self.ipv4_dns_search
+        if self.ipv6_dns_search:
+            retval += " --ipv6-dns-search=%s" % self.ipv4_dns_search
+        if self.ipv4_ignore_auto_dns:
+            retval += " --ipv4_ignore_auto_dns=%s" % self.ipv4_dns_search
```
bcl, line 301: copy paste? :) bools shouldn't have = in the output

pykickstart/commands/network.py
```diff
@@ -824,6 +847,53 @@ def parse(self, args):
...
+                        help="""
+                        Use this option to set IPv4 search domains.
+                        
+                        Requires ``--device`` to be specified.""")
+        op.add_argument("--ipv6-dns-search", default=None, version=F38, dest="ipv6_dns_search",
+                        help="""
+                        Use this option to set IPv6 search domains.
+
+                        Requires ``--device`` to be specified.""")
+        op.add_argument("--ipv4-ignore-auto-dns", type=ksboolean, default=None, version=F38,
```
bcl, line 866: Hm, I now see you're using ksboolean here. Ends up we really don't use that much (looks like network --onboot and repo --ignoregroups are the only two). It's probably better to just make it a simple bool.

tests/commands/network.py
```diff
@@ -357,6 +357,31 @@ def runTest(self):
...
+        network_data = self.assert_parse("network --device eth0 --ipv4-dns-search example.com --ipv6-dns-search example.com")
+        self.assertEqual(network_data.ipv4_dns_search, "example.com")
+        self.assertEqual(network_data.ipv6_dns_search, "example.com")
+        network_data = self.assert_parse("network --device eth0")
+        self.assertIsNone(network_data.ipv4_dns_search)
+        self.assertIsNone(network_data.ipv6_dns_search)
+        self.assert_parse_error("network --ipv4-dns-search example.com --ipv6-dns-search example.com")
+
+        # ignore auto dns
+        network_data = self.assert_parse("network --device eth0 --ipv4-ignore-auto-dns yes --ipv6-ignore-auto-dns yes")
```
bcl, line 374: I don't think having yes/no/true/false available really adds anything. The option itself is verbose enough to be understood.

pykickstart/commands/network.py
```diff
@@ -824,6 +847,53 @@ def parse(self, args):
...
         return retval
 
+class F38_Network(F27_Network):
+    removedKeywords = F27_Network.removedKeywords
+    removedAttrs = F27_Network.removedAttrs
+
+    def _getParser(self):
+        op = F27_Network._getParser(self)
+        op.add_argument("--ipv4-dns-search", default=None, version=F38, dest="ipv4_dns_search",
+                        help="""
```
bcl, line 857: May as well add examples to these while you're at it.

VladimirSlavik:
I will ask in the bug, so that there's more certainty for the decisions about the form(at).
- Why `None` - the setting is boolean-ish, but I assumed we want to skip touching anything if this wasn't seen.
Why `ksboolean` - the only precedents for boolean-ish options use that. There is no "non-verbose boolean" option. We could go for the "present=one state, not present=another state" variant, but that precludes the "stay default" option above.

VladimirSlavik:
(Fixed the output name - kickstart itself uses dashes, while the code has underscores - it was mixed up.)

bcl:
Ah! that makes sense, thanks.

VladimirSlavik:
Turns out that we can expect the defaults to stay always the same, so this can be trimmed down to just have the `--disable-***` options. Goodbye, ksboolean revival :'-(

VladimirSlavik:
This is now the simplified variant.

One remaining doubt is how to call the options. `--ipv4-dns-search` might feel incomplete, but `--ipv4-dns-search-domains` becomes quite long.

pykickstart/commands/network.py
```diff
@@ -280,6 +280,29 @@ def _getArgsAsStr(self):
...
+        self.ipv6_ignore_auto_dns = kwargs.get("ipv6_ignore_auto_dns", None)
+
+    def _getArgsAsStr(self):
+        retval = F27_NetworkData._getArgsAsStr(self)
+        if self.ipv4_dns_search:
+            retval += " --ipv4-dns-search=%s" % self.ipv4_dns_search
+        if self.ipv6_dns_search:
+            retval += " --ipv6-dns-search=%s" % self.ipv4_dns_search
+        if self.ipv4_ignore_auto_dns:
+            retval += " --ipv4-ignore-auto-dns %s" % self.ipv4_dns_search
```
bcl, line 301: copy paste typo on these two options :)

tests/commands/network.py
```diff
@@ -357,6 +357,27 @@ def runTest(self):
...
+        network_data = self.assert_parse("network --device eth0 --ipv4-dns-search example.com --ipv6-dns-search example.com")
+        self.assertEqual(network_data.ipv4_dns_search, "example.com")
+        self.assertEqual(network_data.ipv6_dns_search, "example.com")
+        network_data = self.assert_parse("network --device eth0")
+        self.assertIsNone(network_data.ipv4_dns_search)
+        self.assertIsNone(network_data.ipv6_dns_search)
+        self.assert_parse_error("network --ipv4-dns-search example.com --ipv6-dns-search example.com")
+
+        # disable auto dns
+        network_data = self.assert_parse("network --device eth0 --ipv4-ignore-auto-dns --ipv6-ignore-auto-dns")
```
bcl, line 374: Also, why didn't this catch the typo above...

VladimirSlavik:
Sorry, that exactly smart :-/ Fixed, hopefully.

tests/commands/network.py
```diff
@@ -357,6 +357,27 @@ def runTest(self):
         self.assert_parse(vlan_over_bond_cmd)
         self.assert_parse(vlan_over_bond_cmd + " --bindto mac")
 
+class F38_TestCase(F27_TestCase):
+    def runTest(self):
```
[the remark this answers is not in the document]
bcl: Could you add one more? For the case where both are set:

`self.assert_parse("network --device eth0 --ipv4-dns-search example-one.com --ipv6-dns-search example-two.com"")`

Thanks!

VladimirSlavik:
Fixed the tests to check for crossovers thoroughly.

bcl (approved):
Looks good!

pykickstart/commands/network.py
```diff
@@ -824,6 +847,53 @@ def parse(self, args):
...
+class F38_Network(F27_Network):
+    removedKeywords = F27_Network.removedKeywords
+    removedAttrs = F27_Network.removedAttrs
+
+    def _getParser(self):
+        op = F27_Network._getParser(self)
+        op.add_argument("--ipv4-dns-search", default=None, version=F38, dest="ipv4_dns_search",
+                        help="""
+                        Use this option to set IPv4 search domains. For example: ``--ipv4-dns-search domain1.example.com,domain2.example.com``
+
```
rvykydal, line 859: Seems like there are some trailing spaces here?

rvykydal (approved):
Looks good to me, thank you.

pykickstart/commands/network.py
VladimirSlavik: Thanks! Fixed.

VladimirSlavik:
Should this wait for the anaconda PR, os they both go together? There's more to do there.

bcl:
> Should this wait for the anaconda PR, or do they both go together?

If you're pretty sure the pykickstart part is stable then I can go ahead and merge this and build it. Let me know.

VladimirSlavik:
The anaconda part is now acked too, so feel free to merge.

PR #434, opened 2023-02-07
Enable container support for ostreesetup command

jkonecny12:
Add new --bootc parameter which changes behavior of the command.

Related: rhbz#2125655


pykickstart/commands/ostreesetup.py
```diff
@@ -93,3 +97,61 @@ class RHEL8_OSTreeSetup(F21_OSTreeSetup):
...
+
+class F38_OSTreeSetup(F21_OSTreeSetup):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.bootc = kwargs.get('bootc', False)
+
+    def __str__(self):
+        retval = super().__str__()
+
+        if self.bootc:
```
bcl, line 109: This should be added in an overridden `getArgsAsStr` function.

pykickstart/commands/ostreesetup.py
```diff
@@ -93,3 +97,61 @@ class RHEL8_OSTreeSetup(F21_OSTreeSetup):
...
+            retval = retval.strip()
+            retval += " --bootc\n"
+
+        return retval
+
+    def _getParser(self):
+        op = super()._getParser()
```
bcl, line 116: I think we'll have to replace the original _getParser() completely, since this is using the same arguments, but making changing required to false. I think it's going to raise an error if you try to set them again.

pykickstart/commands/ostreesetup.py
```diff
@@ -93,3 +97,61 @@ class RHEL8_OSTreeSetup(F21_OSTreeSetup):
...
+        if self.bootc:
+            if self.remote:
+                raise KickstartParseError(_("--bootc and --remote cannot be used together"),
+                                          lineno=self.lineno)
+            if self.ref:
+                raise KickstartParseError(_("--bootc and --ref cannot be used together"),
+                                          lineno=self.lineno)
+        # Require arguments when bootc is not used
+        else:
+            super()._check_arguments()
```
bcl, line 150: This should use `F21_OSTreeSetup` instead of super, for consistency with the rest of pykickstart.

bcl (changes requested):
I know this is just a draft, just a few comments.

tests/commands/ostreesetup.py
```diff
@@ -70,5 +70,32 @@ class RHEL7_TestCase(F21_TestCase):
     def runTest(self):
         F21_TestCase.runTest(self)
 
+class F38_TestCase(F21_TestCase):
+    def runTest(self):
+        # Check default arguments settings
+        cmd = F38_OSTreeSetup()
```
poncovka, line 76: I think you should move this part to a separate test case.

tests/commands/ostreesetup.py
```diff
@@ -70,5 +70,32 @@ class RHEL7_TestCase(F21_TestCase):
...
+
+        # no remote, should equal osname (check for compatibility)
+        cmd.parse('--osname=fedora-atomic --url=http://example.com/repo --ref=fedora-atomic/sometest/base/core'.split(' '))
+        self.assertEqual(cmd.remote, cmd.osname)
+
+        # check backward compatibility
+        F21_TestCase.runTest(self)
+
+        # pass
+        self.assert_parse("ostreesetup --bootc --url=quay.io/fedora/coreos:stable")
```
poncovka, line 88: This doesn't check the generated string since you didn't specified the expected value for comparison.

poncovka (commented):
After seeing the implementation, I still think it would be better to have a new kickstart command.

bcl:
> WDYT @bcl?

Yeah, I suppose so. I like `ostreecontainer` more than `bootc` as a name because it's pretty obvious to non-ostree people what it is, or is not :) Whereas `bootc` could be confused with normal booting of some kind.

jkonecny12:
Changed the code totally to add a new command instead of the current one.

jkonecny12:
What I'm thinking now is if we want to somehow raise an exception in case `ostreesetup` and `ostreecontainer` commands are used together? If yes, I guess `ostreecontainer` should have preference? Also wonder how can I access other commands in the given command?

bcl:
> What I'm thinking now is if we want to somehow raise an exception in case `ostreesetup` and `ostreecontainer` commands are used together? If yes, I guess `ostreecontainer` should have preference? Also wonder how can I access other commands in the given command?

Anaconda will have to do that check -- I don't think we have any mechanism in pykickstart that could do that. But yes, something should check and return an error. I wouldn't give preference to either, just raise an error and let the user sort it out.

poncovka:
> Anaconda will have to do that check -- I don't think we have any mechanism in pykickstart that could do that. But yes, something should check and return an error. I wouldn't give preference to either, just raise an error and let the user sort it out.

Actually, pykickstart is able to check that. See the [autopart command](https://github.com/pykickstart/pykickstart/blob/21ee3c413771aa40d352a05c9994a0bfe72c55a2/pykickstart/commands/autopart.py#L197:L213) or just search `commands can't be used at the same time`. It is just not a very convenient. I was going to open an issue about that, because I am working on some new commands for the Stratis support and it will be a nightmare to implement these checks in the current way.

To clarify the issue, since the checks are done in the `parse` methods, it is necessary to implement them in all conflicting commands, because you can only verify commands that were parsed before the currently parsed command. You have no information about the commands that are going to be parsed.

For example, the conflict between `autopart` and `raid` will be reported by the `raid` command in this case:

```
autopart
raid
```

However, it will be reported by the `autopart` command in this case:

```
raid
autopart
```

bcl:
Right, so wouldn't it be more straightforward to do the check in Anaconda after the whole kickstart has been parsed? Although that does mean ksvalidator can't catch the problem :/

jkonecny12:
*UPDATED:*
* Fix the command.
* Fix `seen` variable which was missing resulting in generating both ostree commands in Anaconda side.
* Implement conflicts between `ostreesetup` and `ostreecontainer`.

I decided to add the conflicting part because it will fail early so it should provide better UX for end user.

pykickstart/commands/ostreecontainer.py
```diff
@@ -82,4 +85,9 @@ def _getParser(self):
     def parse(self, args):
         ns = self.op.parse_args(args=args, lineno=self.lineno)
         self.set_to_self(ns)
+
+        if self.handler.ostreesetup.seen:
```
jstodola, line 89: Shouldn't it also conflict with `liveimg`? And with `url` and `nfs` and `cdrom` and `harddrive` and `hmc`?
I just wonder if this is the best approach to solve conflicting commands in general.

bcl:
So currently we really don't have any up-front checks. If you run ksvalidator with all the install options it happily says it passes.
So I'm not too worried about trying to solve that in this PR. I'll make a new issue for that and look into solving this in a more general way.

pykickstart/commands/ostreecontainer.py
bcl: oops, put my comment in the main thread, sorry. I think that's valid, but should be handled in a new PR.

bcl (dismissed):
Looks ok to me, thanks!

poncovka (commented):
Looks good to me. Thanks!

jkonecny12:
**UPDATED:**
* Renamed `osname` to `stateroot` as was done by the `ostree` tooling.
* Set default value to `stateroot` as proposed in https://github.com/ostreedev/ostree/issues/2794

@cgwalters do you agree with the command?

jkonecny12:
@cgwalters can you please point us to place with description of URL? I thought that it would be just simple `quay.io/fedora/coreos:stable`, however, after playing with that on system it seems that we need a special format which is not that simple and could be extended. Based on this https://github.com/ostreedev/ostree-rs-ext/blob/b93946a1c512661ed4026914ff48c7c4ef25d18a/lib/src/container/mod.rs#L167

I found out what should be probably used. However, we need a way to describe that in pykickstart documentation and I wasn't able to find source for that. Could you please give us helping hand here?

jkonecny12:
@cgwalters  also I wonder, when I look on help of `ostree` command I found this:

```
        --imgref <IMGREF>
            Source image reference, e.g.
            ostree-remote-image:someremote:registry:quay.io/exampleos/exampleos@sha256:abcd...
```
Does that mean that the `someremote` is our `remote` argument from the command or is that something different?

jkonecny12:
Rebased.

cgwalters:
> @cgwalters can you please point us to place with description of URL? I thought that it would be just simple quay.io/fedora/coreos:stable, however, after playing with that on system it seems that we need a special format which is not that simple and could be extended. 

Right, I think we discussed this in the BZ.  Indeed we invented a special string format which I think was a mistake.  In bootc which is a new "frontend" to this I've avoided it in favor of explicit command-line options.

```
# bootc switch --help
bootc-switch 
Target a new container image reference to boot

USAGE:
    bootc switch [OPTIONS] <TARGET>

ARGS:
    <TARGET>    Target image to use for the next boot

OPTIONS:
    -h, --help
            Print help information

        --no-signature-verification
            Explicitly opt-out of requiring any form of signature verification

        --ostree-remote <OSTREE_REMOTE>
            Enable verification via an ostree remote

        --quiet
            Don't display progress

        --retain
            Retain reference to currently booted image

        --transport <TRANSPORT>
            The transport; e.g. oci, oci-archive.  Defaults to `registry` [default: registry]
```

So I'd say ideally we just clone this flow for the container model.

`ostreecontainer --url=ostree-unverified-registry:quay.io/someuser/exampleos:latest` would be equivalent to `ostreecontainer --no-signature-verification quay.io/someuser/exampleos:latest` etc.

IOW if we can make this CLI basically equivalent to what's in `bootc switch` that would increase alignment here.

cgwalters:
(scrolling up we're going back over https://github.com/pykickstart/pykickstart/pull/434#issuecomment-1422515049 right?)

cgwalters:
BTW for a bit more rationale behind why there's this extra stuff on top of a basic container image reference - it's around integrity, which is important because we're going to *boot* these images.  More in https://github.com/containers/skopeo/issues/1829

jkonecny12:
I talked to @travier today and we had a debugging session. Seems that we are not able to resolve Anaconda blocking issue now. So, I'll finish this and Anaconda PRs and also mark this as experimental (agreed with @cgwalters on that).

The question is, if we get this in after the beta freeze. Before the beta freeze we are probably not able to fix the Anaconda issue so it would be completely unusable if merged. Switching back to draft until the Anaconda issue is resolved.

jkonecny12:
**UPDATED:**
* Squash `stateroot` rename
* Fix tests
* Remove logic to avoid printing default values for `stateroot` and `remote`

IIRC pykickstart did not usually care about printing default values into output kickstart which honestly seems good practice to me because it will make the resulting kickstart even more robust. Just want to make sure that everyone agree with this change.

jkonecny12:
> So I'd say ideally we just clone this flow for the container model.
> 
> `ostreecontainer --url=ostree-unverified-registry:quay.io/someuser/exampleos:latest` would be equivalent to `ostreecontainer --no-signature-verification quay.io/someuser/exampleos:latest` etc.
> 
> IOW if we can make this CLI basically equivalent to what's in `bootc switch` that would increase alignment here.

@cgwalters I don't want to make Anaconda / pykickstart so smart. The underlying tools should be smart and we should just give them data. Based on that I have a few questions.

- Do I understand it correctly that `ostree container image deploy` requires the scheme-reference (`ostree-unverified-registry`)? Based on my testing it does.
- Because this basically breaks my network detection (the url start is completely different). Can I detect network requirement by just testing for the `remote` word in the scheme-reference?
- Is the `--no-signature-verification` even required? I mean the `unverified` keyword is already part of the scheme-reference. What will happen if user do `--no-signature-verification --url=ostree-image-signed:...` isn't it conflicting request?

cgwalters:
On Wed, Feb 22, 2023, at 11:58 AM, Jiri Konecny wrote:
>> So I'd say ideally we just clone this flow for the container model.
>> 
>> `ostreecontainer --url=ostree-unverified-registry:quay.io/someuser/exampleos:latest` would be equivalent to `ostreecontainer --no-signature-verification quay.io/someuser/exampleos:latest` etc.
>> 
>> IOW if we can make this CLI basically equivalent to what's in `bootc switch` that would increase alignment here.
>> 
> @cgwalters <https://github.com/cgwalters> I don't want to make Anaconda 
> / pykickstart so smart. The underlying tools should be smart and we 
> should just give them data. 

I agree!  In the hopefully near future we can rebase this all on bootc instead, which presents a better interface.

> Based on that I have a few questions.
>
>  • Do I understand it correctly that `ostree container image deploy` 
> requires the scheme-reference (`ostree-unverified-registry`)? Based on 
> my testing it does.

Right, yes,

>  • Because this basically breaks my network detection (the url start is 
> completely different). Can I detect network requirement by just testing 
> for the `remote` word in the scheme-reference?

For now…hmm, let’s split by first colon and then use the existing function.  

>  • Is the `--no-signature-verification` even required? I mean the 
> `unverified` keyword is already part of the scheme-reference. What will 
> happen if user do `--no-signature-verification 
> --url=ostree-image-signed:...` isn't it conflicting request?

Right.  Let’s aim to only accept one form of reference here.  Either the ostree one or the bootc-style one.

jkonecny12:
> • Because this basically breaks my network detection (the url start is completely different). Can I detect network requirement by just testing for the `remote` word in the scheme-reference?
> For now…hmm, let’s split by first colon and then use the existing function.

What about `` ostree-remote-image:someremote:registry:quay.io/exampleos/exampleos@sha256:abcd`` form. You have more than just one column for the metadata before actual URL. I mean `someremote` even though, I'm not 100% sure what should go there it seems to be supported. BTW I have that example from `ostree container image deploy --help`.

Because of that I don't think removing just first part before column would help. I would need to also find out about the `someremote` if it is available or not. So I would rather just detect the `remote` word from the first part and avoid parsing anything else.

> • Is the `--no-signature-verification` even required? I mean the `unverified` keyword is already part of the scheme-reference. What will happen if user do `--no-signature-verification --url=ostree-image-signed:...` isn't it conflicting request?
> Right. Let’s aim to only accept one form of reference here. Either the ostree one or the bootc-style one.

In that case I will go with just URL and remove `--no-signature-verification`. The second solution (removing scheme-reference and leaving parameter) would mean pykickstart understands what is supported. That could be pretty fragile.

jkonecny12:
**UPDATED:**
* Add `--transport`
* Simplify `--url`
* Remove default value from `stateroot`

We had another discussion with @cgwalters and the result was these improvements to usability to ostree:
https://github.com/ostreedev/ostree-rs-ext/pull/464
https://github.com/ostreedev/ostree-rs-ext/pull/462

These basically means that we don't need to care about `stateroot` default value but mainly we don't need to care about `reference-scheme` which made lot of issues discussed above. This way I'm adding new `--transport` value and leaving here `--no-signature-verification` which giving the same information as `reference-scheme` but in much more understandable and reliable value. Thank @cgwalters for your help!

pykickstart/commands/ostreecontainer.py
```diff
@@ -0,0 +1,101 @@
...
+
+    def _getParser(self):
+        op = KSOptionParser(prog="ostreecontainer", description="""
+                            Used for OSTree installations from native container. See
+                            https://coreos.github.io/rpm-ostree/container/
+                            for more information about OSTree.
+
+                            **Experimental. Use on your own risk.**
+                            """, version=F38)
+        # Remane the osname to stateroot and set default as proposed by
```
bcl, line 71: `Rename` typo :)

tests/commands/ostreecontainer.py
```diff
@@ -0,0 +1,74 @@
+#
+# Colin Walters <walters@redhat.com>
```
bcl, line 2: Probably don't need this here.

bcl (approved):
Looks ok, see #442 for conflict support that will need to be changed once it's merged.

jkonecny12:
**UPDATED:**
* Fixed notes from @bcl, thanks!
* Add support for the new conflicting commands.

The API seems to be stable from now on. Now it should be just about fixing Anaconda parts where I was able to make first finished installation. Marking this PR as Ready.

jkonecny12:
Please release this change to Fedora 38 too when possible.

PR #435, opened 2023-02-16
Add support for Fedora 39

VladimirSlavik:
Not sure what to write here :)

### DIFF
--- a/pykickstart/handlers/f39.py
+++ b/pykickstart/handlers/f39.py
@@ -1,0 +1,124 @@
+#
+# Copyright 2023 Red Hat, Inc.
+#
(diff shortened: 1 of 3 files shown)


pykickstart/handlers/f39.py
```diff
@@ -0,0 +1,124 @@
+#
+# Copyright 2022 Red Hat, Inc.
```
poncovka, line 2: 2023

poncovka (commented):
Thanks!

pykickstart/handlers/f39.py
VladimirSlavik: Thanks!

PR #438, opened 2023-02-20
Improve local rpm build experience

jkonecny12:
This will allow us to do `make rc-release` and get rpm files of pykickstart. Otherwise it's pretty hard to do when you don't know the project so much.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,6 +13,8 @@ GPGKEY ?= $(shell git config user.signingkey)
 WEBLATE_REPO = git@github.com:pykickstart/weblate
 WEBLATE_BRANCH ?= $(shell git branch --show-current)
 
+SPECFILE ?= pykickstart.spec
+
 tests := $(wildcard tests/*py tests/commands/*py tests/tools/*py)
 
 all:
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -133,12 +135,13 @@ scratch: docs
...
 	@rm -rf /tmp/pykickstart-$(VERSION) /tmp/pykickstart
 	@dir=$$PWD; cp -a $$dir /tmp/pykickstart-$(VERSION)
+	@cp /tmp/pykickstart-$(VERSION)/docs/_build/text/kickstart-docs.txt /tmp/pykickstart-$(VERSION)/docs
 	@cd /tmp/pykickstart-$(VERSION) ; $(PYTHON) setup.py -q sdist
 	@cp /tmp/pykickstart-$(VERSION)/dist/pykickstart-$(VERSION).tar.gz .
 	@rm -rf /tmp/pykickstart-$(VERSION)
 	@echo "The archive is in pykickstart-$(VERSION).tar.gz"
 
-rc-release: scratch-bumpver scratch pykickstart.spec
+rc-release: scratch-bumpver scratch pykickstart.spec sign
```
bcl, line 144: I don't like requiring signing for rc builds. It makes it harder for someone who doesn't have gpg setup, and also increases the changes of accidentally releasing signed code by accident.

In weldr-client I made signing conditional (on by default), we should do something similar here - https://github.com/osbuild/weldr-client/blob/main/weldr-client.spec.in

bcl (changes requested):

Makefile
jkonecny12: Fixed

jkonecny12:
Thanks for merging this! It will make my life a bit better ;)

PR #442, opened 2023-03-01
Add conflicting commands support

bcl:
Previously a command would have to implement checking the .seen attribute on commands that it conflicts with, and manually update the command description with the commands that it could not be used with.

With this change you can implement command conflict support by adding a list to the class:

  conflictingCommands = ["raid", "reqpart"]

Pass this list to the command's KSOptionParser as ``conflicts``:

  KSOptionParser(..., conflicts=self.conflictingCommands)

And then in the first version that the conflict occurs, call the helper function with a translated error message from the ``parse`` function:

  self._checkConflictingCommands(_("The %s and autopart commands can't be used at the same time"))

the command's description will have a list of the conflicting commands appended to it automatically.

### DIFF
--- a/pykickstart/commands/logvol.py
+++ b/pykickstart/commands/logvol.py
@@ -620,9 +620,6 @@ class RHEL6_LogVol(F12_LogVol):
     def parse(self, args):
         # call the overriden method
         retval = F12_LogVol.parse(self, args)
-        # the logvol command can't be used together with the autopart command
-        # due to the hard to debug behavior their combination introduces
-        self._checkConflictingCommands(_("The logvol and %s commands can't be used at the same time"))
 
         if retval.thin_volume and retval.thin_pool:
             errorMsg = _("--thin and --thinpool cannot both be specified for "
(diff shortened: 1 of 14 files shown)


pykickstart/commands/logvol.py
```diff
@@ -621,9 +622,7 @@ def parse(self, args):
         retval = F12_LogVol.parse(self, args)
         # the logvol command can't be used together with the autopart command
         # due to the hard to debug behavior their combination introduces
-        if self.handler.autopart.seen:
-            errorMsg = _("The logvol and autopart commands can't be used at the same time")
-            raise KickstartParseError(errorMsg, lineno=self.lineno)
+        self._checkConflictingCommands(_("The logvol and %s commands can't be used at the same time"))
```
poncovka, line 625: I think this can be automated as well.

poncovka (approved):
I like the solution with `conflictingCommands`, but I would take the automation a little further. See https://github.com/poncovka/pykickstart/commit/6801f188fcff04f9a594d24f78661ba412bd27b9. However, it requires some changes in [unit tests](https://github.com/poncovka/pykickstart/commit/5f63a8fa7b17cb3af9261ecd40d9f4ba37c9fcf0).

Otherwise, it looks good to me. Thank you!

pykickstart/commands/logvol.py
VladimirSlavik: This took me a bit to understand. Do you mean that every command should run this code, so that the check can't be forgotten? That sounds good to me, it will be a noop with empty conflicts anyway.

bcl:
> but I would take the automation a little further.

Thanks! I was avoiding doing that in this PR because I was thinking the error string might need to be customized, but in practice they are all the same for now so that looks good to me.

> I did not entirely understand is how the conflicts get to the docs.

In `pykickstart/options.py` the `KSOptionParser` modifies the command's description string with the list of conflicting commands.

poncovka:
Thanks!

PR #445, opened 2023-03-02
Fix handling of package section arguments in older versions

### LINKED ISSUE #444: KS version not respected
Koji uses pykickstart to read a kickstart template, modify it and write out a new kickstart. pykickstart isn't respecting the version of the original template and it's regenerating a new kickstart that does not comply with the version that was requested.

I wrote a test case to (hopefully) illustrate the error:

```python
from pykickstart.version import RHEL8
class Packages_Options_VersionRespected(ParserTest):
    def __init__(self, *args, **kwargs):
        ParserTest.__init__(self, *args, **kwargs)
        self.ks = """
#version=RHEL8
%packages --instLangs="es_ES" --excludeWeakdeps
%end
"""

    def runTest(self):
        self.parser.readKickstartFromString(self.ks)

        # Verify that the options are parsed as expected
        self.assertTrue(self.handler.packages.excludeWeakdeps)
        self.assertEqual(self.handler.packages.instLangs, "es_ES")

        # And then test that all those options would be printed out
        # correctly.
        self.assertEqual(str(self.handler.version), RHEL8)
        self.assertEqual(str(self.handler.packages), "\n%packages --instLangs=es_ES --excludeWeakdeps\n\n%end\n")
```

The first two asserts pass as expected, but the last two fail. The generated kickstart looks like this:

```
# Generated by pykickstart v3.44
#version=DEVEL
# System bootloader configuration
bootloader --location=none
  
%packages --inst-langs=es_ES --exclude-weakdeps
%end
```

Of course, when this gets passed to a RHEL8 version of anaconda, it complains because `--inst-langs` and `--exclude-weakdeps` are unrecognized arguments.

Note: in my current Koji installation (which uses pykickstart v3.32), the `version` is generated correctly but the package arguments are still wrong. This may just be that my test is wrong, I'm not sure.

bcl:
The %package section argument were changed in F32 to prefer the non-camelcase form. But when using an older version they should be output using the old form, not the new one.

Fixes #444

### DIFF
--- a/pykickstart/parser.py
+++ b/pykickstart/parser.py
@@ -380,11 +380,17 @@ class Packages(KickstartObject):
         if self.handleBroken == constants.KS_BROKEN_IGNORE:
             retval += " --ignorebroken"
         if self.instLangs is not None:
-            retval += " --inst-langs=%s" % self.instLangs
+            if self._ver < version.F32:
+                retval += " --instLangs=%s" % self.instLangs
+            else:
+                retval += " --inst-langs=%s" % self.instLangs
         if self.multiLib:
             retval += " --multilib"
         if self.excludeWeakdeps:
-            retval += " --exclude-weakdeps"
+            if self._ver < version.F32:
+                retval += " --excludeWeakdeps"
+            else:
+                retval += " --exclude-weakdeps"
         if self.timeout is not None:
             retval += " --timeout=%d" % self.timeout
         if self.retries is not None:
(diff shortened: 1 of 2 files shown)


VladimirSlavik (approved):
Thanks for fixing this, and sorry fir the bug!

poncovka (approved):
Thank you!

alexiri:
For what it's worth, we've included this patch in our rebuild of pykickstart and we're happily using it in production.

PR #447, opened 2023-03-15
displaymode: Update description to describe behavior

bcl:
The displaymode commands (text, cmdline, graphical) control the mode during the installation, and the mode that the system will boot into by default. This updates the description to correctly reflect that behavior.


jkonecny12 (approved):
Looks great, thank you.

I wonder if we want to mention how to change the result by `services` command set to graphical? At least I think that is the correct way :D.

Issue #338, opened 2020-09-03
Report that wireless options have no effect


missic:
https://docs.fedoraproject.org/en-US/fedora/rawhide/install-guide/advanced/Boot_Options/#sect-boot-options-removed lists 'essid=, wepkey=, wpakey=' options as removed due to network configuration being handled differently than in the past.
ksvalidator should point this out.
In fact, I have tried to install f32 with those options and the process failed with an error message about "devicewifi" and "get_carrier" (unfortunately I don't have the full error message handy, but removing those options and installing via ethernet cable worked).

bcl:
Those are not kickstart options, those are kernel cmdline options.

missic:
Those are also listed as arguments to the 'network' kickstart command: https://pykickstart.readthedocs.io/en/latest/kickstart-docs.html#network

jkonecny12:
@rvykydal could you please tell us state of this in Anaconda?

bcl:
> Those are also listed as arguments to the 'network' kickstart command: https://pykickstart.readthedocs.io/en/latest/kickstart-docs.html#network

If there is a problem with them in that context please file a bug against Anaconda at https://bugzilla.redhat.com

Woi:
There is already a bug report/feature request: https://bugzilla.redhat.com/show_bug.cgi?id=1927325

Would be nice to have this documented. It took me several hours to find out why it's not working

PR #450, opened 2023-04-08
Document missing WIFI support in Anaconda

Woi:
See #338 for details

### DIFF
--- a/pykickstart/commands/network.py
+++ b/pykickstart/commands/network.py
@@ -451,7 +451,10 @@ class FC3_Network(KickstartCommand):
                         command beyond the first.
                         """)
         op.add_argument("--essid", version=FC3,
-                        help="The network ID for wireless networks.")
+                        help="""
+                        The network ID for wireless networks.
+                        Needs additonal support by installation programm. This 
+                        is currently not the case for anaconda.""")
         op.add_argument("--ethtool", version=FC3, help="""
                         Specifies additional low-level settings for the network
                         device which will be passed to the ethtool program.""")


bcl:
I'm reluctant to start putting warnings like this into kickstart. It looks like this isn't supported in anaconda, and it tells you so. Looking at livecd-creator, it looks like that's probably why this was added.

bcl:
> So where does it tell me so? Maybe we can work on that to improve user experience.

When you try to run anaconda with that included it prints the error. I haven't tried it myself, but the code is here:
https://github.com/rhinstaller/anaconda/blob/master/pyanaconda/modules/network/initialization.py#L93

Oh, and I now see it is logged as info, and it continues running without stopping. I'd take that up with Anaconda -- if it were me I'd prefer that it stops and throws an error since it's too easy to miss in the logs and I assume anyone including that really meant for it to set things up.

Woi:
Thanks. You're absolutely right, I definitely preferred Anaconda to stop and inform me. That would have saved me  hours of debugging my kickstart file, while still missing this particular log message.
So I followed your suggestion and asked the Anaconda folks for a change: https://bugzilla.redhat.com/show_bug.cgi?id=1927325#c5

Woi:
Slightly off topic: as a user the differences between pykickstart and Anaconda is not obvious. For me it even looked as the same in the beginning. Anaconda pointing people that [want to do an automated installation!](https://github.com/rhinstaller/anaconda#i-want-to-do-an-automated-installation) to pykickstart documentation supports this wrong impression.

bcl:
Yeah, it can be a bit confusing at first. I'm not sure of any way to make things clearer -- we don't want to duplicate documentation, and pykickstart supports commands from previous versions as well so it's the best place to point people to instead of duplicating everything.

PR #452, opened 2023-04-08
Fix issues how to generate encrypted passwords

### LINKED ISSUE #451: Documentation: multiple issues in code example for generating encrypted passwords
I found multiple issues using `  ``python -c 'import crypt; print(crypt.crypt("My Password", "$6$My Sault"))'``  ` from [documentation](https://github.com/pykickstart/pykickstart/blob/master/pykickstart/commands/user.py#L179) to create encrypted passwords on Fedora 37 
* First of all: it's not working, due to a space in "My Sault"
* "Sault" is a typo
* Salt is not randomised; people might reuse it due to blindly copy and pasting it
* Password could be readable in shell history
* crypt() is deprecated and slated for removal in Python 3.13
* SHA-512 is used while yescrypt is the [default since Fedora 35](https://fedoraproject.org/wiki/Changes/yescrypt_as_default_hashing_method_for_shadow) and supported [from Fedora 29 and RHEL 9 upwards](https://www.openwall.com/yescrypt/)
* Confusing and unnecessary quotes: \`` at the beginning and at the end.

Woi:
Introduce a better option but keep a fixed version of the old code example for older distributions Fixes #451

### DIFF
--- a/pykickstart/commands/user.py
+++ b/pykickstart/commands/user.py
@@ -174,11 +174,17 @@ class FC6_User(KickstartCommand):
                         password argument is assumed to already be encrypted.
                         ``--plaintext`` has the opposite effect - the password
                         argument is assumed to not be encrypted. To create an
-                        encrypted password you can use python::
+                        encrypted password you can use::
 
-                        ``python -c 'import crypt; print(crypt.crypt("My Password", "$6$My Sault"))'``
+                        mkpasswd -m yescrypt
+                        
+                        This will generate a yescrypt hash of your password using
+                        a random salt. As a fallback for older distributions or
+                        in case mkpasswd is not available you can use::
+                        
+                        python -c 'import crypt; print(crypt.crypt("My Password", "$6$MySalt"))'
 
-                        This will generate sha512 crypt of your password using
+                        This will generate a SHA-512 hash of your password using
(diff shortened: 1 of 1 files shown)


pykickstart/commands/user.py
```diff
@@ -174,11 +174,17 @@ def _getParser(self):
                         password argument is assumed to already be encrypted.
                         ``--plaintext`` has the opposite effect - the password
                         argument is assumed to not be encrypted. To create an
-                        encrypted password you can use python::
+                        encrypted password you can use::
 
-                        ``python -c 'import crypt; print(crypt.crypt("My Password", "$6$My Sault"))'``
+                        mkpasswd -m yescrypt
```
bcl, line 179: This should be `mkpasswd -m sha512crypt` for consistency.

bcl (changes requested):
Thanks, just change yescrypt to sha512crypt and it's good.

pykickstart/commands/user.py
bcl: Somehow I missed we'd changed to yescrypt. Nevermind :)

bcl (approved):
Looks good, thanks!

Woi:
Nice, thanks :)

PR #455, opened 2023-05-01
workflow: Update actions to newest versions

bcl:


bcl:
Can't get upstream coveralls action to work, so I only merged the checkout and python-setup action changes.

PR #456, opened 2023-05-15
realm: switch from pipes.quote() to shlex.quote()

ptoscano:
The 'pipes' module was marked deprecated in Python 3.11, slated for removal in Python 3.13 [1]. The only bit of it used was `pipes.quote()`, which can be easily replaced by `shlex.quote()`, as it is also implemented in the 'pipes' module.

[1] https://peps.python.org/pep-0594/#pipes


VladimirSlavik (approved):
Looks good to me, thank you!

PR #459, opened 2023-08-08
rtd: formats is separate, not part of sphinx

bcl:


bcl:
I just love not being able to test 3rd party service config files before pushing. So let's play whack a bug!

PR #460, opened 2023-08-08
Fix some python 3.12 deprecations and add testing on py3.12

bcl:


VladimirSlavik:
Are you using the same translation canary as we do? Maybe you want to sync it? We fixed it for python 3.12 earlier with blivet... https://github.com/rhinstaller/translation-canary/pull/7/files

bcl:
> Are you using the same translation canary as we do?

Updated it to pull it in as a subtree, thanks. It may have been initialized by doing that but wasn't clear from the logs.

Also note that pylint will fail, the current version (2.17.5) doesn't support python 3.12 so it crashes when trying to analyze the `docs/versionremoved.py` code.

VladimirSlavik:
In anaconda, we sort of gave up on pylint for now, and rely on ruff...

VladimirSlavik:
I agree, pylint cannot be fully replaced. If not for the pre-production Python in Fedora, we'd be running it same as we used to.

bcl:
https://bugzilla.redhat.com/show_bug.cgi?id=2246138 rawhide failure when using py3.13

PR #461, opened 2023-10-24
Add ostreecontainer command

kwozyman:
Backport of new ostreecontainer command to RHEL9

### DIFF
--- a/pykickstart/commands/ostreecontainer.py
+++ b/pykickstart/commands/ostreecontainer.py
@@ -1,0 +1,94 @@
+#
+# Copyright (C) 2023  Red Hat, Inc.
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# the GNU General Public License v.2, or (at your option) any later version.
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY expressed or implied, including the implied warranties of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General
+# Public License for more details.  You should have received a copy of the
+# GNU General Public License along with this program; if not, write to the
+# Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+# 02110-1301, USA.  Any Red Hat trademarks that are incorporated in the
+# source code or documentation are not subject to the GNU General Public
+# License and may only be used or replicated with the express permission of
+# Red Hat, Inc.
+#
+from pykickstart.version import F38
+from pykickstart.base import KickstartCommand
+from pykickstart.options import KSOptionParser
+
+class F38_OSTreeContainer(KickstartCommand):
+    removedKeywords = KickstartCommand.removedKeywords
+    removedAttrs = KickstartCommand.removedAttrs
+
+    def __init__(self, *args, **kwargs):
+        KickstartCommand.__init__(self, *args, **kwargs)
+        self.op = self._getParser()
+        self.stateroot = kwargs.get('stateroot', None)
+        self.url = kwargs.get('url', None)
+        self.transport = kwargs.get("transport", None)
+        self.remote = kwargs.get("remote", self.stateroot)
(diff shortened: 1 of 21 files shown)


pykickstart/commands/ostreecontainer.py
```diff
@@ -0,0 +1,94 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General
+# Public License for more details.  You should have received a copy of the
+# GNU General Public License along with this program; if not, write to the
+# Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+# 02110-1301, USA.  Any Red Hat trademarks that are incorporated in the
+# source code or documentation are not subject to the GNU General Public
+# License and may only be used or replicated with the express permission of
+# Red Hat, Inc.
+#
+from pykickstart.version import F38
```
bcl, line 18: This should not refer to F38 which from the RHEL9 perspective is in the future :) Just change them to RHEL9.

pykickstart/commands/ostreesetup.py
```diff
@@ -93,3 +95,8 @@ class RHEL8_OSTreeSetup(F21_OSTreeSetup):
 
 class RHEL9_OSTreeSetup(F21_OSTreeSetup):
     pass
+
+class F38_OSTreeSetup(F21_OSTreeSetup):
```
bcl, line 99: Same here, this can just become part of the `RHEL9_OSTreeSetup` class.

pykickstart/version.py
```diff
@@ -93,6 +93,7 @@
 F32 = 32000
 F33 = 33000
 F34 = 34000
+F38 = 38000
```
bcl, line 96: No need for this.

pykickstart/version.py
```diff
@@ -107,7 +108,7 @@
     "F19": F19, "F20": F20, "F21": F21, "F22": F22, "F23": F23,
     "F24": F24, "F25": F25, "F26": F26, "F27": F27, "F28": F28,
     "F29": F29, "F30": F30, "F31": F31, "F32": F32, "F33": F33,
-    "F34": F34,
+    "F34": F34, "F38": F38,
```
bcl, line 111: Or for this.

tests/commands/ostreecontainer.py
```diff
@@ -0,0 +1,59 @@
...
+        cmd = F38_OSTreeContainer()
+        self.assertEqual(cmd.noSignatureVerification, False)
+
+        # test if arguments are required
+        op = cmd._getParser()
+        for action in op._actions:
+            if '--url' in action.option_strings:
+                self.assertEqual(action.required, True)
+
+class F38_TestCase(CommandTest):
```
bcl, line 34: Name this RHEL9_

tests/commands/ostreesetup.py
```diff
@@ -70,5 +71,17 @@ class RHEL7_TestCase(F21_TestCase):
     def runTest(self):
         F21_TestCase.runTest(self)
 
+class F38_Conflict_TestCase(CommandSequenceTest):
```
bcl, line 74: RHEL9_ here as well

bcl (changes requested):
Other than those, this looks ok.

pykickstart/commands/reqpart.py
```diff
@@ -19,14 +19,12 @@
 #
 from pykickstart.version import F23
 from pykickstart.base import KickstartCommand
-from pykickstart.errors import KickstartParseError
 from pykickstart.options import KSOptionParser
 
-from pykickstart.i18n import _
```
bcl, line 25: This needs to remain :)

bcl: Actually, no, the later commit that adds back the message needs to be changed. Maybe these were applied in the wrong order?

VladimirSlavik: Looking at #434 where this should be ported from, I'm similarly confused - the commit order seems to be reversed, and some commits missing.

PR #467, opened 2023-11-09
remove control characters from license

### LINKED ISSUE #465: pip install of pykickstart doesn't install requests dependency
The current `pyproject.toml` is incorrect, the dynamic entries have to point to files, as described in the documentation here - https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#dynamic-metadata so that may be why `requests` isn't installed. But I have been unable to fix it.

I have tried:
* Adding a simple `dependencies` section with `requests` in it.
* Adding a new requirements.pip.txt file with `requests` in it.
* Removing the `install_requires` in setup.py as a diagnostic (tox needs it)

I am testing it by running (from within a py3.11 venv): `python3 -I -m pip install -vvv --check-build-dependencies -I .` and it installs pykickstart into the venv, but totally ignores any kind of dependency I add -- including totally bogus ones.

supakeen:
Having control characters in the LICENSE file prevents setuptools from installing dependencies when used in PEP-517 compliant mode. See: https://github.com/pypa/setuptools/issues/4033

I spent entirely too much time on this. This closes #465.

### DIFF
--- a/COPYING
+++ b/COPYING
@@ -55,7 +55,7 @@ patent must be licensed for everyone's free use or not licensed at all.
 
   The precise terms and conditions for copying, distribution and
 modification follow.
-
+
 		    GNU GENERAL PUBLIC LICENSE
    TERMS AND CONDITIONS FOR COPYING, DISTRIBUTION AND MODIFICATION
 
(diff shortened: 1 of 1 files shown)


VladimirSlavik (approved):
Looks good to me. Thank you!

PR #468, opened 2023-11-09
Add Fedora 40 support

bcl:


VladimirSlavik (approved):
I can't remember if this was everything, but we'd find soon enough I think :)

PR #471, opened 2024-01-30
ostreecontainer: Add --target-transport and --target-url

cgwalters:
These will be combined and plumbed down into the existing ostree CLI which has `--target-imgref`.


cgwalters:
(Not tested for anything beyond the unit tests)

pykickstart/commands/ostreecontainer.py
```diff
@@ -77,6 +83,12 @@ def _getParser(self):
         op.add_argument("--transport", version=F38, help="""
                         The transport; e.g. registry, oci, oci-archive.
                         The default is `registry`.""")
+        op.add_argument("--target-url", version=F38, help="""
```
bcl, line 86: version refers to the Fedora version of Anaconda this command is added to, is it going to be backported all the way to F38 or is this going to be F40?

cgwalters: Yeah changed to use 40

cgwalters:
> Do you have a corresponding Anaconda change?

nope

PR #478, opened 2024-03-27
github: Limit python version tested based on PR branch

bcl:
This PR attempts to limit which python versions are used to test the branches. eg. it makes no sense to test rhel8 against python 3.12 or to test rhel10-branch against anything lower than python 3.12


bcl:
No idea how to make this work, or if it is even possible.

PR #482, opened 2024-04-02
Add the missing rhsm and syspurpose commands

bcl:


bcl:
Same as #481 just for the rhel10-branch

PR #485, opened 2024-04-18
Add a DEFAULT_VERSION variable to version.py

bcl:
This is meant to make it easier to branch for RHEL versions where DEVEL isn't the right thing to default to. DEFAULT_VERSION isn't visible to the user, it is used to set the expected default when nothing else is passed -- usually this is the same as DEVEL. But on RHEL we can set this to be RHEL10 instead, switching it to the right parser with a 1 line change.


rvykydal (approved):
Looks good to me.

PR #491, opened 2024-05-06
Deprecate modularity on Fedora 41

marusak:


jkonecny12:
I'm not sure now but I have a feeling that we shouldn't inherit RHEL classes from Fedora ones. @bcl will know better than me.

marusak:
https://github.com/pykickstart/pykickstart/pull/491/files#diff-174e6c3287219cc2642021f34a456021f50e447b2064536584fde6c9fba7975cR73 Same thing is happening here

bcl:
Usually we want to avoid doing that, changes should always be upstream first when possible but looking at older commits I see we haven't always done that.

I think this is fine as long as we don't make it a habit :)

pykickstart/commands/module.py
```diff
@@ -188,3 +188,7 @@ def _getParser(self):
 
         """ % versionToLongString(RHEL10))
         return op
+
+class F41_Module(RHEL10_Module):
```
bcl, line 192: This can't inherit from RHEL10, if it does then the deprecation message is going to be wrong.
RHEL messages can reference Fedora since they are based on them, but Fedora cannot reference RHEL, it needs to say it was deprecated in Fedora 41.

bcl (changes requested):

marusak:
Fair points, so I change it around. Now F41 is the base where modularity is deprecated and RHEL10 just inherits it.

bcl (approved):
Thanks!

PR #495, opened 2024-07-03
commands: Add missing DeprecatedCommand class to F41_Module

KKoukiou:
Also increase consistently with other deprecated commands, by calling the DeprecatedCommand init, instead of the custom parser.

### DIFF
--- a/tests/commands/module.py
+++ b/tests/commands/module.py
@@ -21,8 +21,7 @@
 
 import unittest
 from tests.baseclass import CommandTest
-from pykickstart.errors import KickstartDeprecationWarning
-
+from pykickstart.base import DeprecatedCommand
 
 class F29_TestCase(CommandTest):
     def runTest(self):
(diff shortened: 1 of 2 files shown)


KKoukiou:
I just noticed that the module is missing the DeprecatedCommand. I am not sure where it's used, but I today [switched some of Anaconda tests](https://github.com/rhinstaller/anaconda/pull/5741) to rely on RemovedCommand for some skipping. So let's be correct.

tests/commands/module.py
```diff
@@ -77,9 +77,13 @@ def runTest(self):
 
 class F41_TestCase(F31_TestCase):
     def runTest(self):
-        # F41 warns about deprecation
-        with self.assertWarns(KickstartDeprecationWarning):
-            F31_TestCase.runTest(self)
+        # make sure we've been deprecated
+        parser = self.getParser("module")
+        self.assertEqual(issubclass(parser.__class__, DeprecatedCommand), True)
```
bcl, line 82: You need to import `DeprecatedCommand` and drop the `KickstartDeprecationWarning` import.

bcl (changes requested):

bcl (approved):
Looks good, thanks!

PR #496, opened 2024-07-19
pykickstart is supported on python 3.6+

sgallagher:
This is to make the metadata consistent with reality (the tests all pass on python 3.6)


bcl (approved):
Looks good to me, thanks.

PR #497, opened 2024-07-25
Mark the vnc command as deprecated on RHEL 10

M4rtinK:
VNC has been replaced by the Remote Desktop protocol, which is currently controlled just by the inst.rdp boot option and does not have a corresponding
kickstart command.

Resolves: RHEL-41219

### DIFF
--- a/pykickstart/commands/vnc.py
+++ b/pykickstart/commands/vnc.py
@@ -17,8 +17,8 @@
 # subject to the GNU General Public License and may only be used or replicated
 # with the express permission of Red Hat, Inc.
 #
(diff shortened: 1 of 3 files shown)


KKoukiou (approved):

bcl (approved):
Looks ok to me, need to do the same thing on master for the rhel10 class there as well.

M4rtinK:
> Looks ok to me, need to do the same thing on master for the rhel10 class there as well.

Yep, here you go: https://github.com/pykickstart/pykickstart/pull/498 :)

PR #498, opened 2024-07-26
Mark the vnc command as deprecated on RHEL 10

M4rtinK:
VNC has been replaced by the Remote Desktop protocol, which is currently controlled just by the inst.rdp boot option and does not have a corresponding
kickstart command.

(cherry picked from commit 34fafc1d658dd44c99c914f615194621c4f2874e)

Resolves: RHEL-41219

Upstream version of: https://github.com/pykickstart/pykickstart/pull/497


bcl (approved):
Looks good, thanks!

PR #499, opened 2024-08-05
[WIP] add support for home reuse to autopart

rvykydal:


pykickstart/commands/autopart.py
```diff
@@ -598,6 +598,69 @@ def parse(self, args):
...
+                        version=F41, help="""
+                        Specifies which existing mountpoins should be
+                        reused.
+
+                        Single device must exist for given mountpoint.
+
+                        ``autopart --reuse=/home``""")
+        op.add_argument("--erase", dest="erase", type=commaSplit,
+                        version=F41, help="""
+                        Specifies which existing mountpoins should be
```
bcl, line 639: I would assume this is non-recursive, right? May want to make a note of that just to be clear that erasing / doesn't also erase /home if it is a separate mountpoint.

bcl:
I'm not opposed to this, but I also wonder if there is another way? Would it instead make it clearer if `part` were used to set the behavior for the mountpoints? Also, how does it identify the previous mounting locations of various devices reliably?

The failed tests were unrelated, I've just fixed those on master so a rebase will take care of them.

pykickstart/commands/autopart.py
```diff
@@ -598,6 +598,69 @@ def parse(self, args):
...
+            retval += " --remove=" + ",".join(self.remove)
+
+            retval += "\n"
+        return retval
+
+    def _getParser(self):
+        op = super()._getParser()
+        op.add_argument("--reuse", dest="reuse", type=commaSplit,
+                        version=F41, help="""
+                        Specifies which existing mountpoins should be
```
Conan-Kudo, line 631: `mountpoins` -> `mountpoints`

Conan-Kudo (changes requested):

rvykydal:
> I'm not opposed to this, but I also wonder if there is another way? Would it instead make it clearer if `part` were used to set the behavior for the mountpoints? Also, how does it identify the previous mounting locations of various devices reliably?

The approach is described in https://github.com/rhinstaller/anaconda/pull/5814
It is using *ExistingSystems DeviceTree API to find existing installations and therefore mount points. And automatic partitioning request structure to define what to do (pass reuse/erasse/remove requests to the autopartitioning).
Our (initial?) use case is limited to single Linux system (perhaps even single system).
It is possible that we would support multiple systems and selection in the UI. Then passing mount points would be ambigous, so we'd need to pass respective (selected) devices to the backend - in that case this kickstart api would really make less sense. (We could maybe allow --reuse /home:DEVNAME then, but it would require more thinking). Overall it seems to me for more complicated / ambigous use cases users should rather use mount point assignment. We should just offer simple use case valuable for most of the users here.

PR #502, opened 2024-08-13
Add "opal-admin-passphrase" option for autopart and part commands

vojtechtrefny:
This enables support for creating LUKS HW-OPAL devices in Anaconda The feature itself is controlled by the existing "--luks-version" option, but we need the OPAL administrator passphrase from user.

### DIFF
--- a/pykickstart/commands/autopart.py
+++ b/pykickstart/commands/autopart.py
@@ -19,7 +19,7 @@
 #
 from pykickstart.base import KickstartCommand
 from pykickstart.version import versionToLongString, RHEL6, RHEL7, RHEL8
-from pykickstart.version import FC3, F9, F12, F16, F17, F18, F20, F21, F26, F29, F38
+from pykickstart.version import FC3, F9, F12, F16, F17, F18, F20, F21, F26, F29, F38, F41
 from pykickstart.constants import AUTOPART_TYPE_BTRFS, AUTOPART_TYPE_LVM, AUTOPART_TYPE_LVM_THINP, AUTOPART_TYPE_PLAIN
 from pykickstart.errors import KickstartParseError
 from pykickstart.options import KSOptionParser
@@ -630,3 +630,38 @@ class RHEL10_AutoPart(F38_AutoPart):
 
                     Partitioning scheme 'btrfs' was removed.""" % versionToLongString(RHEL8)
         return op
+
+class F41_AutoPart(F38_AutoPart):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.hw_passphrase = kwargs.get("hw_passphrase", "")
+
+    def __str__(self):
+        retval = super().__str__()
+        if not self.autopart:
+            return retval
+
+        if self.encrypted and self.hw_passphrase:
+            retval = retval.strip()
+            retval += " --hw-passphrase=\"%s\"" % self.hw_passphrase
+            retval += "\n"
+        return retval
+
+    def _getParser(self):
+        op = super()._getParser()
+        op.add_argument("--hw-passphrase", dest="hw_passphrase", version=F41, default="",
+                        help="""
+                        Only relevant if ``--encrypted`` is specified and ``--luks-version``
+                        is set to either `luks2-hw-opal` or `luks2-hw-opal-only`.
+
+                        OPAL administrator passphrase needed to create a new OPAL locking range.""")
+        return op
+
+    def parse(self, args):
+        retval = super().parse(args)
+
+        if self.hw_passphrase and (not self.encrypted or not self.luks_version.startswith("luks2-")):
+            msg = _("hw-passphrase may be specified only with LUKSv2 HW-OPAL encrypted devices")
+            raise KickstartParseError(msg, lineno=self.lineno)
+
+        return retval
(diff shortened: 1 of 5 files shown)


bcl:
Do we really need a new argument for this? We already have `--passphrase` on both of those. The only reason I could see for adding a new option is if you needed to specify more than one passphrase type, and I don't think that's the case here, is it?

vojtechtrefny:
> Do we really need a new argument for this? We already have `--passphrase` on both of those. The only reason I could see for adding a new option is if you needed to specify more than one passphrase type, and I don't think that's the case here, is it?

We actually need two different passphrases -- one for the LUKS keyslot (same as with normal LUKS) and the OPAL administrator passphrase which is used when creating or managing the hardware encryption and these can be different so we cannot use just one option.

vojtechtrefny:
I now realize I didn't put both `--opal-admin-passphrase` and `--passphrase` in the test cases so it might not be obvious that these are two different passphrases and both are needed. Normally user would specify both. (And I guess `--passphrase` is not used in other test cases because it is not strictly required, Anaconda can ask for LUKS passphrase when not specified in kickstart.)

bcl:
Oh boy, ok. Last question, I think, does it need to be named opal-admin? Or would a more generic name like `--hw-passphrase` sound good? I'm trying to anticipate things like adding other non-opal hardware support.

vojtechtrefny:
I doubt we'll support anything else than OPAL in the future, but I agree it's better to have something future proof. I changed the name to `--hw-encryption-passphrase`, it's quite long but I think having just `--hw-passphrase` might be confusing.

bcl:
> I doubt we'll support anything else than OPAL in the future, but I agree it's better to have something future proof. I changed the name to `--hw-encryption-passphrase`, it's quite long but I think having just `--hw-passphrase` might be confusing.

Why would that be confusing? 'passphrase' already implies encryption so I think it's redundant to add that to it. And it makes it quite long :)

vojtechtrefny:
> Why would that be confusing? 'passphrase' already implies encryption so I think it's redundant to add that to it. And it makes it quite long :)

When I see "HW passphrase" the first thing that comes to my mind is BIOS passphrase, so not necessary encryption. We could call it "SED passphrase", that's what sedutil calls it. But I always confuse SED and the util `sed`. Maybe I am overthinking this a little too much.

bcl:
I think as long as we document it `--hw-passphrase` is a good compromise. If it was for the bios I'd call it `--bios-passphrase` :)

vojtechtrefny:
> I think as long as we document it `--hw-passphrase` is a good compromise. 

Ok, done. And with that I think this is ready for a review. I've also updated the Anaconda PR with the new argument name.

pykickstart/commands/autopart.py
```diff
@@ -630,3 +630,38 @@ def _getParser(self):
...
+                        is set to either `luks2-hw-opal` or `luks2-hw-opal-only`.
+
+                        OPAL administrator passphrase needed to create a new OPAL locking range.""")
+        return op
+
+    def parse(self, args):
+        retval = super().parse(args)
+
+        if self.hw_passphrase and (not self.encrypted or not self.luks_version.startswith("luks2-")):
+            msg = _("OPAL admin passphrase may be specified only with LUKSv2 HW-OPAL encrypted devices")
```
bcl, line 664: Change `OPAL admin passphrase` to `hw-passphrase` so it's clearer what part of the arguments it is talking about.

pykickstart/commands/partition.py
```diff
@@ -779,3 +792,24 @@ def _getParser(self):
...
+                        is set to either `luks2-hw-opal` or `luks2-hw-opal-only`.
+
+                        OPAL administrator passphrase needed to create a new OPAL locking range.""")
+        return op
+
+    def parse(self, args):
+        retval = super().parse(args)
+
+        if retval.hw_passphrase and (not retval.encrypted or not retval.luks_version.startswith("luks2-")):
+            msg = _("OPAL admin passphrase may be specified only with LUKSv2 HW-OPAL encrypted devices")
```
bcl, line 812: Same here, call it `hw-passphrase`

tests/commands/autopart.py
```diff
@@ -360,5 +360,13 @@ def runTest(self):
         F38_TestCase.runTest(self)
         self.assert_parse_error("autopart --type=btrfs")
 
+class F41_TestCase(F38_TestCase):
+    def runTest(self):
+        F38_TestCase.runTest(self)
+        self.assert_parse("autopart --encrypted --luks-version=luks2-hw-opal --hw-passphrase=\"test\"",
```
bcl, line 366: Also test with `--hw-passphrase` and `--passphrase` as different strings to ensure they aren't cross-populated.

tests/commands/partition.py
```diff
@@ -366,5 +366,15 @@ def  runTest(self):
         F29_TestCase.runTest(self)
         self.assert_parse_error("part / --fstype=btrfs")
 
+class F41_TestCase(F34_TestCase):
+    def runTest(self):
+        F34_TestCase.runTest(self)
+
+        self.assert_parse("part / --encrypted --luks-version=luks2-hw-opal --hw-passphrase=\"test\"",
```
bcl, line 373: Same here, add a test with both passphrases.

bcl (changes requested):
A couple small changes, and I think the commit message also needs to be updated.

pykickstart/commands/autopart.py
vojtechtrefny: Done.

bcl (approved):
Thanks!

PR #503, opened 2024-08-15
Enhance %onerror section with recommendation

jkonecny12:
Add information that %onerror could be forced by inst.cmdline.

Resolves: RHEL-47142

Not 100% sure if we want to mention Anaconda kernel parameter in pykickstart documentation but I see this beneficial.


jkonecny12:
Also take the resolves as pointer only. For RHEL we will update RHEL documentation (issue is filed).

bcl (approved):
Looks good to me.

PR #504, opened 2024-08-19
DeprecatedCommand: Return empty list for dataList

bcl:
Returning None can cause callers, like anaconda, to traceback when they expect to get a list. This was revealed by the deprecated module command. Also include a test for this.


jkonecny12:
Thank you for a quick fix!

PR #506, opened 2024-09-10
Deprecate VNC on Fedora 42 (also add Fedora 42)

jkonecny12:
**Please do not merge until changes in Anaconda are merged**

We are switching to RDP, currently, without pykickstart support.

https://fedoraproject.org/wiki/Changes/Anaconda_As_Native_Wayland_Application

=================

I'm not sure if this is correct implementation or I should inherit from the rhel-10 class. Feel free to say preference @bcl .

Also a note that there are some voices from community that they want to keep VNC with GRD. However, to my knowledge VNC implementation in GRD is not great and not recommended by GRD developer.


pykickstart/commands/vnc.py
```diff
@@ -149,3 +149,12 @@ def _getParser(self):
         op = F9_Vnc._getParser(self)
         op.description += "\nPlease use the ``inst.rdp`` boot option instead.\n\n.. deprecated:: %s" % versionToLongString(RHEL10)
         return op
+
+class F42_Vnc(DeprecatedCommand, F9_Vnc):
```
bcl, line 153: I think it's ok to just subclass RHEL10_Vnc here instead of duplicating things.

tests/commands/vnc.py
```diff
@@ -70,5 +70,11 @@ def runTest(self):
         parser = self.getParser("vnc")
         self.assertEqual(issubclass(parser.__class__, DeprecatedCommand), True)
 
+class F42_TestCase(F9_TestCase):
+    def runTest(self):
+        # make sure the vnc command has been deprecated on RHEL 10
```
bcl, line 75: Comment typo :) should be 'Fedora 42' not 'RHEL 10'

bcl:
Also if you want to split out the F42 version switch into a separate PR we can get that in first since it's past branching time and F42 now exists.

jkonecny12:
The F42 handler was split from this PR. I'll rebase this when https://github.com/pykickstart/pykickstart/pull/507 is merged.

jkonecny12:
I've updated the commit to fix the RHEL-10 leftover and try to base it on RHEL10_Vnc class. However, it won't help much because I still want to change the documentation for deprecation in F42 instead of RHEL-10 in Fedora. At least I guess that is correct approach.

bcl:
> I still want to change the documentation for deprecation in F42

Ah, right, if it's different then change whatever is needed.

pykickstart/commands/vnc.py
jkonecny12: Done.

jkonecny12:
This should have been in Fedora 42 already so let's make it ready for review. I guess I forgot about his PR, sorry.

PR #508, opened 2024-10-02
options: adjust to behavior change in upstream _parse_optional

AdamWill:
In Python 3.13 and 3.12.7, the behavior of _parse_optional has changed. It used to raise an error on multiple matching actions itself, and only ever return None or an option tuple. Now the "raise error on multiple matching actions" code was moved out into consume_optional, and _parse_optional returns either None or a *list* of option tuples, which contains more than one if multiple actions match. See:

https://github.com/python/cpython/pull/124631
https://github.com/python/cpython/issues/58573

This adapts to the change in a way that should work on both older and newer Pythons. I re-implemented the handling of multiple matching options so we don't have to worry about licensing and attribution if we copied in upstream's version.

### DIFF
--- a/pykickstart/options.py
+++ b/pykickstart/options.py
@@ -177,9 +177,23 @@ class KSOptionParser(ArgumentParser):
         self.lineno = None
 
     def _parse_optional(self, arg_string):
-        option_tuple = ArgumentParser._parse_optional(self, arg_string)
+        # Before 3.13 and 3.12.7, this returned None or a single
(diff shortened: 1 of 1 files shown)


AdamWill:
btw, AFAICS, the only purpose of this wrapper for upstream `_parse_optional` is to allow the printing of a much more specific message for deprecated options. perhaps we could look at sending a patch for argparse upstream to allow custom deprecation messages?

PR #510, opened 2024-10-03
Fix the fix for _parse_optional changing

AdamWill:
Whoops. Don't know how this slipped through testing.


bcl (approved):
Ack. argh. ugh. linting (automated and eyeball) fail.

PR #514, opened 2024-10-04
Add support for inline certificates with `%certificate` section

KKoukiou:
**Please do not merge until changes in Anaconda are implemented as well**

This commit introduces a new `%certificate` section, allowing users to securely embed certificates directly within the kickstart file.
The section also supports an optional `--path` argument, specifying where the certificate should be installed on the target system.

Example:

%certificate --name custom_certificate.pem --path /etc/pki/ca-trust/source/anchors/ -----BEGIN CERTIFICATE-----
MIIDkDCCAnigAwIBAgIUFfTKU02DB4Nz4u3pRk1ShpRvn0AwDQYJKoZIhvcNAQEL ... -----END CERTIFICATE-----
%end

### DIFF
--- a/pykickstart/parser.py
+++ b/pykickstart/parser.py
@@ -195,6 +195,9 @@ class Certificate(KickstartObject):
 
     def __str__(self):
         """Return a string formatted for output to a kickstart file."""
+        if not self.cert:
+            return ""
+
         retval = "\n%certificate"
 
         retval += " --name=%s" % self.name
--- a/pykickstart/parser.py
+++ b/pykickstart/parser.py
@@ -24,6 +24,8 @@ Main kickstart file processing module.
 
 This module exports several important classes:
 
+    Certificate - Representation of a single %certificate section.
+
     Script - Representation of a single %pre, %pre-install, %post, or %traceback script.
 
     Packages - Representation of the %packages section.
@@ -42,7 +44,8 @@ from pykickstart.errors import KickstartError, KickstartParseError, KickstartPar
 from pykickstart.ko import KickstartObject
 from pykickstart.load import load_to_str
 from pykickstart.options import KSOptionParser
-from pykickstart.sections import PackageSection, PreScriptSection, PreInstallScriptSection, \
+from pykickstart.sections import CertificateSection, PackageSection, \
+                                 PreScriptSection, PreInstallScriptSection, \
                                  PostScriptSection, TracebackScriptSection, OnErrorScriptSection, \
                                  NullSection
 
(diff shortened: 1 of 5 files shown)


bcl:
Looks good so far. From the discussions I have seen one of the requirements was to also be able to install them on the target system, I was thinking an argument like `--install` would be good for that, like we do for the `repo` command. We might also need a second path for the target system? But I would hope not.

Could also take a look at the parser.py and base.py coverage, it looks like it fell a little. Coveralls isn't cooperating this morning so I'm not sure exactly which lines, but you can look at the output of `make test` to see the missed lines.

jkonecny12:
Let's start simple and always copy the certificates into the installed system on the same place as we will install them in initrd. If someone asks for not installing them then we can extend that.

bcl:
A test to make sure not passing `--name` generates an error would also be good to have.

pykickstart/parser.py
```diff
@@ -173,6 +176,38 @@ def next(self):
...
+    """A class representing a single %certificate section."""
+    def __init__(self, *args, **kwargs):
+        """Create a new Certificate instance.  Instance attributes:
+
+           :keyword cert: The certificate to be used.
+           :keyword name: The name of the certificate.
+           :keyword path: The path where the certificate should be stored.
+        """
+        KickstartObject.__init__(self, *args, **kwargs)
+        self.cert = "\n".join(kwargs.get("cert", []))
```
rvykydal, line 192: @KKoukiou wouldn't it be more simple to expect string here and join the lines before creating the object in the section?

pykickstart/sections.py
```diff
@@ -849,3 +849,82 @@ def _warn_alias_future_deprecation(self, option, new_option, lineno):
...
+                            HhcNMjEwNzI2MDg0NDIxWhcNMjIwNzI2MDg0NDIxWjBVMQswCQYDVQQGEwJVUzEb
+                            MBkGA1UECBMSU29tZSBTdGF0ZSBvciBQcm92aW5jZTESMBAGA1UEBxMJUm9hbGQg
+                            T25lMRUwEwYDVQQKEwxFeGFtcGxlIE9yZyBObzAwggEiMA0GCSqGSIb3DQEBAQUA
+                            A4IBDwAwggEKAoIBAQC4/SDsn8RQk0Euh6ZTKq5/Mz34K6QlnrxmAF7B8QGbDiK6
+                            ...
+                            -----END CERTIFICATE-----
+                            %end
+                            """, version=F42)
+
+        op.add_argument("--name", dest="name", required=True, version=F42,
```
jstodola, line 886: Would it make sense to rename `--name` to `--filename`? It would be easier to understand the purpose of the option and avoid the users to use it as a description of the certificate, for example: `--name="Cert for my HTTPS server"`.

pykickstart/sections.py
```diff
@@ -849,3 +849,82 @@ def _warn_alias_future_deprecation(self, option, new_option, lineno):
...
+                        help="""The name of the certificate file.""")
+
+        op.add_argument("--path", dest="path", default=None, version=F42, help="""
```
jstodola, line 889: Would it make sense to rename `--path` to `--dir`? A path could be understood as the whole path to the certificate, like `/etc/pki/ca-trust/source/anchors/my_cert.pem`, `--dir` is more clear, IMO. Also, `--dir` is already used by other kickstart commands (`harddrive` and `nfs`).

rvykydal:
I am going to use `--dir` and `--filename`. I'll open an updated PR soonish.

rvykydal:
Taking over and updating the PR here: https://github.com/pykickstart/pykickstart/pull/517

PR #515, opened 2024-11-04
Support computer-name option for realm command (realm.py)

dsnpl:
Permit the `--computer-name` option to the `realm join` command. Used to allow joining where the computer object name does not match the hostname e.g. when directory serves multiple domains and the fqdn needs to be used as the object name as the hostname alone may not be unique.

The `--computer-name` was introduced into realmd in v0.16.3 in 2016 (see https://gitlab.freedesktop.org/realmd/realmd/-/tags/0.16.3).

### DIFF
--- a/pykickstart/commands/realm.py
+++ b/pykickstart/commands/realm.py
@@ -57,7 +57,8 @@ class F19_Realm(KickstartCommand):
                                                        "membership-software=",
                                                        "one-time-password=",
                                                        "no-password",
-                                                       "computer-ou="))
+                                                       "computer-ou=",
+                                                       "computer-name="))
         except getopt.GetoptError as ex:
             raise KickstartParseError(_("Invalid realm arguments: %s") % ex, lineno=self.lineno)
 


bcl (changes requested):
This needs to be added in a F42 class, and it should have a test for it.
I'd also like to see what @M4rtinK has to say.

dsnpl:
Thanks very much for your comments both - I'm learning a lot more about this!

The anaconda installer looks like it takes the join options but doesn't validate them? 
https://github.com/rhinstaller/anaconda/blob/840d6a43fce702403df0d1ddf33c16fd857d5b45/pyanaconda/modules/security/installation.py#L354

I've tried searching that repo for terms like 'computer-ou' but see no references to it?

M4rtinK:
I'll take a look what we actually do, it has been a while since I looked at this code (like, 10 years probably :P).

pykickstart/commands/realm.py
```diff
@@ -57,7 +58,8 @@ def _parseJoin(self, args):
                                                        "membership-software=",
                                                        "one-time-password=",
                                                        "no-password",
-                                                       "computer-ou="))
+                                                       "computer-ou=",
+                                                       "computer-name="))
```
bcl, line 62: This should not be added to the F19 class.

pykickstart/commands/realm.py
```diff
@@ -101,3 +103,37 @@ def parse(self, args):
...
+
+
+class F42_Realm(F19_Realm):
+    removedKeywords = KickstartCommand.removedKeywords
+    removedAttrs = KickstartCommand.removedAttrs
+
+    def _parseJoin(self, args):
+        try:
+            # We only support these args
+            opts, remaining = getopt.getopt(args, "", ("client-software=",
```
bcl, line 115: There may be a nicer way to do this than repeating the whole function. I'm undecided how much it matters at this point (but if there is ANOTHER change later on then I'd like to see a common function with args passed to it for each version). It's up to you if you want to try to make this more generic and share it between the F19 and F42 classes.

tests/commands/realm.py
```diff
@@ -63,5 +63,16 @@ def runTest(self):
...
+class F42_TestCase(F19_TestCase):
+    def runTest(self):
+        F19_TestCase.runTest(self)
+
+        # Support the --computer-name option
+        self.assert_parse("""realm join --computer-name=blee blah2""")
+
+        # Support the --computer-name option alongside another supported argument 
+        self.assert_parse("""realm join --computer-name=blee --computer-ou=OU=blah blah2""")
+
```
bcl, line 75: Could you add a test for the error case if nothing is passed to it. eg. `realm join --computer-name`

bcl (changes requested):
A few small comments, I'm mostly waiting on @M4rtinK and approval from the Anaconda team to add this.

pykickstart/commands/realm.py
dsnpl: Whoops!!! Thanks.

pykickstart/commands/realm.py
dsnpl: Sounds sensible - I've had a go at this and also added 'os-name' and 'os-version' arguments should someone want to set them (see 0.16.3) https://gitlab.freedesktop.org/realmd/realmd/-/tags.

tests/commands/realm.py
```diff
@@ -73,6 +73,11 @@ def runTest(self):
         # Support the --computer-name option alongside another supported argument 
         self.assert_parse("""realm join --computer-name=blee --computer-ou=OU=blah blah2""")
 
+        # fail - computer name must have argument
+        self.assert_parse_error("""realm join --computer-name=""")
```
dsnpl, line 77: whoops this is wrong - no realm specified.

tests/commands/realm.py
dsnpl: Yup - I've added added this, but it failed to produce an error with an empty argument so I added in a [check](https://github.com/dsnpl/pykickstart/blob/15b04f5a1d4d9d8ca7eaf24807d4a36d184410a4/pykickstart/commands/realm.py#L127) to ensure that a non empty argument was present for `--computer-name`. The default behaviour of getopt is not to error in this case.

Should I add this check for all options that require an argument? If so should I extend this to F19 as well?

Thanks.

pykickstart/commands/realm.py
dsnpl: I'll come back to this once I've worked out how to sufficiently solve https://github.com/pykickstart/pykickstart/pull/515#discussion_r1884360700

tests/commands/realm.py
bcl: Oh, ha. Surprised nobody noticed that before. I'd check all options but just add it to F42, we try not to change behavior in old releases (especially EOL ones) unless there is a really good reason.

PR #517, opened 2024-12-10
%certificates section support

rvykydal:
This is slightly updated https://github.com/pykickstart/pykickstart/pull/514.
The modifications should be clear from the commits, especially it renames the options based on review in the https://github.com/pykickstart/pykickstart/pull/514#pullrequestreview-2478460112

https://issues.redhat.com/browse/INSTALLER-4029

### DIFF
--- a/pykickstart/parser.py
+++ b/pykickstart/parser.py
@@ -24,6 +24,8 @@ Main kickstart file processing module.
 
 This module exports several important classes:
 
+    Certificate - Representation of a single %certificate section.
+
     Script - Representation of a single %pre, %pre-install, %post, or %traceback script.
 
     Packages - Representation of the %packages section.
@@ -42,7 +44,8 @@ from pykickstart.errors import KickstartError, KickstartParseError, KickstartPar
 from pykickstart.ko import KickstartObject
 from pykickstart.load import load_to_str
 from pykickstart.options import KSOptionParser
-from pykickstart.sections import PackageSection, PreScriptSection, PreInstallScriptSection, \
+from pykickstart.sections import CertificateSection, PackageSection, \
+                                 PreScriptSection, PreInstallScriptSection, \
                                  PostScriptSection, TracebackScriptSection, OnErrorScriptSection, \
                                  NullSection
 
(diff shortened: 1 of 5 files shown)


pykickstart/parser.py
```diff
@@ -189,7 +189,7 @@ def __init__(self, *args, **kwargs):
            :keyword path: The path where the certificate should be stored.
         """
         KickstartObject.__init__(self, *args, **kwargs)
-        self.cert = "\n".join(kwargs.get("cert", []))
+        self.cert = "".join(kwargs.get("cert", [])).removesuffix("\n")
```
bcl: This requires python 3.9 which is fine unless we need to backport to RHEL 8. I've modified the github tests in #518 so that we test on 3.9 and later.

tests/certificate.py
```diff
@@ -16,6 +16,25 @@
 ...
 -----END CERTIFICATE-----"""
 
+CERT_CONTENT_WITH_EMPTY_LINE="""-----BEGIN CERTIFICATE-----
```
bcl, line 19: Is this supposed to support multiple certificates per section? For some reason I thought it was only one per section. Which is fine, as long as whatever consumes the .pem file is ok with that.

bcl (changes requested):
This looks pretty good to me. There are a couple commit messages mentioning 'edns' which seems like a typo of certificates :) And I'd like to see these commits squashed so that the new feature is just one commit now that all the details have been nailed down.

pykickstart/parser.py
rvykydal: There are only tickets for rhel 9 and rhel 10 as of now but maybe I'd rather stay on safer side and use something supported.

tests/certificate.py
rvykydal: Yes, I thought so as well.
But it shows up that at this point it is very likely there will be just single a bundle of certs (well-known path and name) that should be managed by user (documented). So no tooling to add certs to a bundle. Therefore Anaconda should rather be able to pass such a bundle via kickstart. And can't think of any other sane solution then being able to dump multiple certs to a single file (for example merge/append in case of the same target -  that should the tooling do, not Anaconda). Also I think the kickstart itself / parser should rather not impose restrictions on the content at this stage of maturity of the feature (actually it is almost just a dump of string).

jkonecny12: Just to be clear. It will allow this section to be used multiple times but also multiple certificates in one section. Am I correct?

jkonecny12 (approved):
I like this new version and I agree with squashing the commits. It's great for re-review but it gives unnecessary dead-ends from history PoV.

pykickstart/parser.py
rvykydal: Updated.

bcl (approved):
Looks good!

rvykydal:
@bcl should I try to make the coveralls failing check green?

bcl:
No, I'm not particularly happy with coveralls right now and want to find a better solution for coverage.

PR #521, opened 2025-01-15
Add support for inline certificates with `%certificate` section

rvykydal:
Port of https://github.com/pykickstart/pykickstart/pull/517


bcl (approved):
Looks good!

PR #536, opened 2025-04-10
Add support for new bootc kickstart command

elkoniu:
Bootc container becomes a new way of OS install supported by the Anaconda. This requires introduction of new kickstart command to support it.

Related discussion: https://github.com/rhinstaller/anaconda/discussions/5197
Resolves: INSTALLER-4024

This is the base for the following change in Anaconda: https://github.com/rhinstaller/anaconda/pull/6298

### DIFF
--- a/pykickstart/commands/__init__.py
+++ b/pykickstart/commands/__init__.py
@@ -22,6 +22,7 @@ from pykickstart.commands import (
     authselect,
     autopart,
     autostep,
+    bootc,
     bootloader,
     btrfs,
     clearpart,
--- a/pykickstart/commands/bootc.py
+++ b/pykickstart/commands/bootc.py
@@ -1,0 +1,100 @@
+#
+# Copyright (C) 2025  Red Hat, Inc.
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# the GNU General Public License v.2, or (at your option) any later version.
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY expressed or implied, including the implied warranties of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General
+# Public License for more details.  You should have received a copy of the
+# GNU General Public License along with this program; if not, write to the
+# Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+# 02110-1301, USA.  Any Red Hat trademarks that are incorporated in the
+# source code or documentation are not subject to the GNU General Public
+# License and may only be used or replicated with the express permission of
+# Red Hat, Inc.
+#
+from pykickstart.version import F43
+from pykickstart.base import KickstartCommand
+from pykickstart.options import KSOptionParser
+
+class F43_Bootc(KickstartCommand):
+    removedKeywords = KickstartCommand.removedKeywords
+    removedAttrs = KickstartCommand.removedAttrs
+    conflictingCommands = ["ostreesetup", "ostreecontainer"]
+
+    def __init__(self, *args, **kwargs):
+        KickstartCommand.__init__(self, *args, **kwargs)
+
+        self.op = self._getParser()
+
+        self.stateroot = kwargs.get('stateroot', None)
+        self.sourceImgRef = kwargs.get("sourceImgRef", None)
+        self.targetImgRef = kwargs.get("targetImgRef", self.sourceImgRef)
+
+    def __str__(self):
+        retval = KickstartCommand.__str__(self)
+
+        if not self.seen:
+            return retval
+
+        retval += "# Bootc setup\n"
+        retval += "bootc %s\n" % self._getArgsAsStr()
+
+        return retval
+
+    def _getArgsAsStr(self):
+        retcmd = []
+
+        if self.stateroot:
+            retcmd.append('--stateroot="%s"' % self.stateroot)
+        if self.sourceImgRef:
+            retcmd.append('--source-imgref="%s"' % self.sourceImgRef)
+        if self.targetImgRef:
+            retcmd.append('--target-imgref="%s"' % self.targetImgRef)
+
+        return ' '.join(retcmd)
+
+    def _getParser(self):
+        op = KSOptionParser(prog="bootc", description="""
+                            Used for Bootc installations from native container. See
+                            https://bootc-dev.github.io/bootc//man/bootc-install-to-filesystem.html
+                            for more information about Bootc install to filesystem.
+
+                            **Experimental. Use on your own risk.**
+                            """, version=F43, conflicts=self.conflictingCommands)
+        # Rename the osname to stateroot and set default as proposed by
+        # https://github.com/ostreedev/ostree/issues/2794
+        op.add_argument("--stateroot",
+                        version=F43,
+                        default="default",
+                        help="""
+                        Name for the state directory, also known as "osname".
+                        """)
+        op.add_argument("--source-imgref",
+                        dest="sourceImgRef",
+                        required=True,
+                        version=F43,
+                        help="""
+                        Install the system from an explicitly given source.
+                        """)
+        op.add_argument("--target-imgref",
+                        dest="targetImgRef",
+                        version=F43,
+                        help="""
+                        Specify the image to fetch for subsequent updates.
+                        If not presented defaults to '--source-imgref' value.
+                        """)
+
+        return op
+
+    def parse(self, args):
+        ns = self.op.parse_args(args=args, lineno=self.lineno)
+        self.set_to_self(ns)
+
+        if not self.targetImgRef:
+            self.targetImgRef = self.sourceImgRef
+
+        return self
+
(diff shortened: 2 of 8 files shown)


bcl:
I realize this is still a draft :) But a few comments:
* This should be for F43, not F41, unless there are plans to backport this. We generally don't do that in Fedora.
* You don't need placeholder classes for RHEL versions that this won't show up in, like RHEL7 and 8.
* You don't need RHEL classes unless they are different, including commands is controlled by the `pykickstart/handlers/` files.

elkoniu:
> I realize this is still a draft :) But a few comments:
> 

Hi Brian, sory for the false alarm - the work is still ongoing :)

>     * This should be for F43, not F41, unless there are plans to backport this. We generally don't do that in Fedora.

Noted and adjusted

> 
>     * You don't need placeholder classes for RHEL versions that this won't show up in, like RHEL7 and 8.
> 
>     * You don't need RHEL classes unless they are different, including commands is controlled by the `pykickstart/handlers/` files.
> 

RHEL related stuff has been removed

> 
> ETA: And there's a typo in the subject line :)

Thanks, fixed

I will continue with draft PR as we are still trying to figure out right set of options for this command. After this will be ready I will ping you and @jkonecny12 for the formal review :)

pykickstart/commands/bootc.py
```diff
@@ -0,0 +1,96 @@
...
+
+                            **Experimental. Use on your own risk.**
+                            """, version=F43, conflicts=self.conflictingCommands)
+        # Rename the osname to stateroot and set default as proposed by
+        # https://github.com/ostreedev/ostree/issues/2794
+        op.add_argument("--stateroot",
+                        version=F43,
+                        help="""
+                        Name for the state directory, also known as "osname".
+                        """)
```
jkonecny12, line 73: Please set the `default` as the default value (based on the man page).

pykickstart/commands/bootc.py
```diff
@@ -0,0 +1,96 @@
...
+        op.add_argument("--source-imgref",
+                        required=True,
+                        version=F43,
+                        help="""
+                        Install the system from an explicitly given source.
+                        """)
+        op.add_argument("--target-imgref",
+                        version=F43,
+                        help="""
+                        Specify the image to fetch for subsequent updates.
```
jkonecny12, line 83: Please add the note that default is using `--source-imgref`.

pykickstart/handlers/f43.py
```diff
@@ -68,6 +68,7 @@ class F43Handler(BaseHandler):
         "timesource": commands.timesource.F33_Timesource,
         "ostreecontainer": commands.ostreecontainer.F38_OSTreeContainer,
         "ostreesetup": commands.ostreesetup.F38_OSTreeSetup,
+        "bootc": commands.bootc.F43_Bootc,
```
jkonecny12, line 71: The list is sorted alphabetically, please move the `bootc` command.

pykickstart/commands/__init__.py
```diff
@@ -64,6 +64,7 @@
     timesource,
     ostreecontainer,
     ostreesetup,
+    bootc,
```
jkonecny12, line 67: The same here. It should be sorted alphabetically.

jkonecny12:
Also please fix the tests.

jkonecny12:
And please fix the typo `fot` in the commit message :)

jkonecny12:
Hi @cgwalters, could you please verify bootc kickstart command API? We tried to get it close to bootc as possible so it is less confusing to move between the tools. We also removed the `remote` as we were not able to find it in `bootc`, not sure if that is a correct approach.

cgwalters (approved):
Looks sane to me superficially. Most other options (e.g. `--karg`) are already wrapped by other anaconda verbs.

One thing not covered though is `--bound-images`; but we probably just want to hardcode `--bound-images=pull` for anaconda, it's effectively what the code is doing now anyways.

bcl:
I agree with @cgwalters, mostly :) If the command is just being run by anaconda then passing through arguments makes sense. If anaconda needs to take some action, or check the validity of arguments before running it then it makes sense to implement them in kickstart. It's a bit of a balance, but the difference in my mind is whether the command is implemented in anaconda, eg. like the part command.

But also remember that there is some value to implementing arguments, it makes it so that ksvalidator can be used on the kickstart to discover errors before actually using it.

cgwalters:
> But also remember that there is some value to implementing arguments, it makes it so that ksvalidator can be used on the kickstart to discover errors before actually using it.

Though bootc is in Rust with clap which is based on a declarative description of the CLI arguments, so in theory we could (like bash completion) generate metadata sufficient to bundle with ksvalidator to do validation there.

(The same applies to a lot of osbuild stage wrappers)

But I dunno, in practice yeah it's nice to be able to validate a kickstart sure, but I think there's so many things that aren't possible to validate that in practice everyone needs to have a good testing framework anyways.

bcl:
Yeah, I don't want to go down that road.

pykickstart/commands/bootc.py
elkoniu: Done

pykickstart/commands/bootc.py
elkoniu: Done

tests/commands/bootc.py
```diff
@@ -0,0 +1,66 @@
+#
+# Copyright 2023 Red Hat, Inc.
```
jkonecny12, line 2: Outdated year.

jkonecny12 (approved):
Looks good to me now. Thanks!

tests/commands/bootc.py
elkoniu: Fixed

jkonecny12:
> Note in the future as bootc evolves I do expect more things to land here, especially [bootc-dev/bootc#22](https://github.com/bootc-dev/bootc/issues/22)
> 
> So one thing I think that would help a lot here is to support a direct passthrough of additional CLI arguments, so we don't need to wrap each individual one in kickstart in the future.
> 
> Prior art for that model looks like https://pykickstart.readthedocs.io/en/latest/kickstart-docs.html#authselect I think?

I like this idea quite a lot, however, I'm afraid how many ways of broken installations we would add this way. For that reason, I would rather implement it on demand honestly. @cgwalters, are you able to tell how fragile the passthrough argument would be?

elkoniu:
@bcl I think we have a merge candidate here :)

tests/commands/bootc.py
```diff
@@ -0,0 +1,66 @@
...
+        for action in op._actions:
+            if '--source-imgref' in action.option_strings:
+                self.assertEqual(action.required, True)
+
+class F43_TestCase(CommandTest):
+    command = "bootc"
+
+    def runTest(self):
+        # PASS tests
+        cmdstr = "bootc --source-imgref=quay.io/centos-bootc/centos-bootc:stream9"
```
bcl, line 39: When writing PASS tests you should also test the output to make sure it is what you expect. In this case the output is different because you are always filling in the target if it isn't set. So it needs to:
* test the output string
* test with setting a different `--target-imgref`

pykickstart/commands/ostreecontainer.py
```diff
@@ -96,3 +96,8 @@ def parse(self, args):
 
 class RHEL9_OSTreeContainer(F38_OSTreeContainer):
     pass
+
+class F43_OSTreeContainer(F38_OSTreeContainer):
+    removedKeywords = KickstartCommand.removedKeywords
+    removedAttrs = KickstartCommand.removedAttrs
+    conflictingCommands = ["ostreesetup", "bootc"]
```
bcl, line 103: Please add bootc to the ostreecontainer conflicting test

pykickstart/commands/ostreesetup.py
```diff
@@ -100,3 +101,8 @@ class F38_OSTreeSetup(F21_OSTreeSetup):
     removedKeywords = KickstartCommand.removedKeywords
     removedAttrs = KickstartCommand.removedAttrs
     conflictingCommands = ["ostreecontainer"]
+
+class F43_OSTreeSetup(F38_OSTreeSetup):
+    removedKeywords = KickstartCommand.removedKeywords
+    removedAttrs = KickstartCommand.removedAttrs
+    conflictingCommands = ["ostreecontainer", "bootc"]
```
bcl, line 108: Please add bootc to the ostreesetup conflict test

bcl (changes requested):

pykickstart/commands/ostreecontainer.py
elkoniu: Done

pykickstart/commands/ostreesetup.py
elkoniu: Done

tests/commands/bootc.py
elkoniu: I have added test you required but from some reason it is failing :) I am wondering if this is some string formatting bug.

bcl: Typo in the assert :) `self.assert_parse(cmdstr, cmdstr)` instead of `self.assert_parse(cmdstr, cmdstr_expected)`
Also still needs a test for using different source and target to demonstrate it not getting copied.

elkoniu: Thanks for pointing this out, I have updated the code following you suggestions.

bcl (approved):
Looks good. I'm not sure why coveralls is reporting twice, but it doesn't matter.

PR #545, opened 2025-06-08
Add support for RDP

jkonecny12:
We removed VNC support recently and replaced that with RDP. We missed the kickstart support for RDP before so now let's fix it and add it back based on the feedback we have got.

### DIFF
--- a/pykickstart/commands/rdp.py
+++ b/pykickstart/commands/rdp.py
@@ -1,0 +1,71 @@
+#
+# Copyright 2025 Red Hat, Inc.
+#
+# This copyrighted material is made available to anyone wishing to use, modify,
+# copy, or redistribute it subject to the terms and conditions of the GNU
+# General Public License v.2.  This program is distributed in the hope that it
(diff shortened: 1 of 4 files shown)


pykickstart/commands/rdp.py
```diff
@@ -0,0 +1,71 @@
...
+                            text installs. With no options, this
+                            command will enable RDP mode in Anaconda and wait
+                            for user to locally set username and password.""",
+                            version=F43)
+        op.add_argument("--username", version=F43, help="""
+                        Set a username which must be provided to connect by
+                        the RPD client.""")
+        op.add_argument("--password", version=F43, help="""
+                        Set a password which must be provided to connect by
+                        the RPD client.""")
```
Conan-Kudo, line 64: `RPD` -> `RDP`

Conan-Kudo (changes requested):

bcl (approved):
Looks good, thanks!

PR #549, opened 2025-07-20
spec: drop python3-build dependency

yselkowitz:
python3-build is not available in the RHEL buildroot and is used only for make targets which are not used when building the package.


Conan-Kudo:
Yeah, `python3-build` will need to be exported in CRB at least going forward.

bcl:
Ah, but in this case I only need it to build the release so it can be removed from the specfile. Installing only requires `python3-pip`

I do expect that ELN is going to need it at some point though.

bcl (approved):
Ack

yselkowitz:
Perhaps, we'll see: https://github.com/fedora-eln/eln/issues/278

PR #553, opened 2025-08-26
Add support for RDP

adamkankovsky:
We removed VNC support recently and replaced that with RDP. We missed the kickstart support for RDP before so now let's fix it and add it back based on the feedback we have got.

Related: RHEL-98217

### DIFF
--- a/pykickstart/commands/__init__.py
+++ b/pykickstart/commands/__init__.py
@@ -66,6 +66,7 @@ from pykickstart.commands import (
     ostreesetup,
     partition,
     raid,
+    rdp,
     reqpart,
     realm,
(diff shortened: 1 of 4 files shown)


M4rtinK (approved):
Looks good to me code wise, but I think there is a typo when linking to the Jira issue. I think it should have this instead:

Resolves: RHEL-96217

(IIRC Related: is not processed by the build system/Jira atomation & this is the correct Jira issue ID for this change.)

M4rtinK:
As agreed with @adamkankovsky I am taking over the PR & I have fixed the issue I have noted during my review (the Jira issue id). :-)

PR #555, opened 2025-11-07
rhsm: Add Flatpak registry option for RHEL10

bruno-fs:
Add --flatpak-registry-url option to configure Flatpak registry settings during installation.

Related: INSTALLER-4482


bcl (approved):
Looks good to me, will also need to be backported to rhel10-branch with the appropriate RHEL-XXXX jira issue.

PR #564, opened 2026-03-24
Fix handling files with special characters

### LINKED ISSUE #8993: All templates and default qubes missing from fresh installation (only dom0 exists) - pykickstart missing quotes when dir has spaces and quotes
### Qubes OS release
4.2, 4.1, 4.1.1

### Brief summary
Unable to install the needed/default templates and the default Vms. No templates in the additional software page during installation.

### Steps to reproduce
Flash Os iso on usb drive
Install the Os on the target machine
Run the Os

### Expected behavior
All the default templates and Vms are installed.

### Actual behavior
No default templates are available. At the first boot, after the post-installation phase, the system is stuck with dom0 and only dom0.

### Detailed explanation

First of all, I’m sorry for my broken english.
I’m here to address an issue that seems to affect very few people, but also seems to be disruptive and insormontable.

I specify that I’m a Qubes user since 2017: I used this OS during the 2017-2019 period, then back again on it in 2022, and today I’m trying to reinstall it.

The issue is that, doesn’t matter what procedure I follow to flash the iso and install the OS, I’m unable to install the needed templates and the default Vms. I’ll leave you some pics that testify how the issue present itself:

First of all, here are the bios settings on my main machine.
photo_5980909691494055883_y
photo_5980909691494055882_y
photo_5980909691494055881_y
photo_5980909691494055880_y

As usual, I flash the iso on the USB drive (tried dd, rufus, etcher: more on that later) and verify the media and the gpg signatures. Then, after the boot, you can see the first symptoms of the issue:
photo_5980909691494055884_y

As you can see, in the software selection page, the templates are totally absent. I can only install the Qubes UI.

I also tried to test the installation media, it seems ok.
photo_5980909691494055886_y

At this point, if I proceed, the installer executes itself and bring me to the reboot. After the reboot, in the post-installation phase, this is the first thing that seems wrong:
photo_5980909691494055887_y

Whonix not available, fedora and debian not even mentioned, and all the options (except the advanced configuration options) are greyed out.

If I go further, in the terminal post-installation phase, the systems gives back this message:
photo_5980909691494055885_y

At this point I can boot Qubes and log in, but I’m stuck with Dom0 and only Dom0. No templates at all, and no pre-configured Vms. I know that there is probably some workaround to this issue, like installing the templates after the Os installation, but I’m not an expert user.

On the forum, a guy told me to run this command in the dom0 terminal, so I leave you also this picture.
photo_5980909691494055888_y

To better address the problem, I tell you that I tried all the possible combinations between the following alternatives (I’ve been on it for the past 3 days):

Downloading the iso:
Tried both from the main qubes website, from the torrent and from multiple mirrors. Tried to flash the 4.2 stable, 4.1 stable and 4.1.1 rc5 (the last time I installed qubes, in 2022, I flashed and installed this last iso and it worked flawlessly). Downloaded the iso both in linux and windows environment.

Flashing the iso:
Tried rufus (both DD and ISO mode), etcher and the dd command in linux environment. Everytime I flashed the iso in a linux environment I did also all the integrity and signature checks, and everytime they passed

Target machine:
The bios settings you see are from the machine with I always used qubes in the past. Anyway, I tried to change the target machine 3 more times:

ASUS UX305UA-FB004T Zenbook, i7-6500U, 8GB Ram
HUAWEI MateBook X Pro, i7-8550U, 8GB Ram
Lenovo IdeaPad S145, i3-8145U, 8GB Ram
Everytime I got the exact same result

Miscellaneous:
Changed 5 different USB drives, also tried a USB 2.0 drive. Tried all the different ports on all the different machines: nothing changed.

Last thing: As you can see here, this isn’t really the first time I encountered this issue. But the last time, in 2022, I was able to resolve it by downloading the 4.1.1 iso (at that time it was a beta version, since the stable was the 4.0). Here you can find the discussion: [All templates MISSING from fresh installation](https://forum.qubes-os.org/t/all-templates-missing-from-fresh-installation/12517)
That brings me to think that there is something wrong with Qubes iso, and not with me.

Link to forum discussion:

https://forum.qubes-os.org/t/are-all-qubes-iso-corrupted-no-additional-software-no-templates-no-default-vms-stuck-with-only-dom0/24722

ben-grande:
It is not always possible to guarantee that the user wrote the media with the method the distribution provided. Enforce quoting to avoid parsing errors from missing quotation end and recognizing an value with spaces as another argument.

Example:

        harddrive --dir=/OS ISO/torrent iso's/Qubes-R4.3.0-x86_64/Qubes-R4.3.0-x86_64.iso --partition=sda2

Fixes: https://github.com/QubesOS/qubes-issues/issues/8993

---

About the choice of single quotes instead of double quotes: some places uses one or the other, I just chose single quotes to avoid making the diff bigger having to escape double quotes or change quote order.

### DIFF
--- a/tests/commands/harddrive.py
+++ b/tests/commands/harddrive.py
@@ -52,20 +52,31 @@ class HardDrive_TestCase(unittest.TestCase):
                        data1, data2,
                        ['biospart', 'partition', 'dir'])
 
+        complicated_dir = """/OS ISO/dir iso's/the.iso"""
+        expected_dir = "--dir=\"%s\"" % complicated_dir
+        data3 = FC3_HardDrive(dir=complicated_dir)
+        data3.seen = True
+        line3 = [line[len("harddrive "):] for line in str(data3).splitlines() if line.startswith("harddrive ")][0]
+        assert expected_dir in line3
+        data4 = FC3_HardDrive(dir=complicated_dir, biospart="sda1")
+        data4.seen = True
+        line4 = [line[len("harddrive "):] for line in str(data4).splitlines() if line.startswith("harddrive ")][0]
+        assert expected_dir in line4
+
 
 class FC3_TestCase(CommandTest):
     def runTest(self):
         # pass
-        self.assert_parse("harddrive --dir=/install --biospart=part", "harddrive --dir=/install --biospart=part\n")
-        self.assert_parse("harddrive --dir=/install --partition=part", "harddrive --dir=/install --partition=part\n")
+        self.assert_parse("harddrive --dir=/install --biospart=part", "harddrive --dir=\"/install\" --biospart=part\n")
+        self.assert_parse("harddrive --dir=/install --partition=part", "harddrive --dir=\"/install\" --partition=part\n")
 
-        self.assertFalse(self.assert_parse("harddrive --dir=/install --partition=sda1") is None)
-        self.assertTrue(self.assert_parse("harddrive --dir=/install --partition=sda1") !=
-                        self.assert_parse("harddrive --dir=/install --partition=sda2"))
-        self.assertFalse(self.assert_parse("harddrive --dir=/install --biospart=80p1") ==
-                         self.assert_parse("harddrive --dir=/install --biospart=80p2"))
-        self.assertFalse(self.assert_parse("harddrive --dir=/install --biospart=sda1") ==
-                         self.assert_parse("harddrive --dir=/other-install --biospart=sda1"))
+        self.assertFalse(self.assert_parse("harddrive --dir=\"/install\" --partition=sda1") is None)
+        self.assertTrue(self.assert_parse("harddrive --dir=\"/install\" --partition=sda1") !=
+                        self.assert_parse("harddrive --dir=\"/install\" --partition=sda2"))
+        self.assertFalse(self.assert_parse("harddrive --dir=\"/install\" --biospart=80p1") ==
+                         self.assert_parse("harddrive --dir=\"/install\" --biospart=80p2"))
+        self.assertFalse(self.assert_parse("harddrive --dir=\"/install\" --biospart=sda1") ==
+                         self.assert_parse("harddrive --dir=\"/other-install\" --biospart=sda1"))
 
         # fail
         # required option --dir missing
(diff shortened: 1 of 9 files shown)


bcl (changes requested):
I don't think we need to resort to shelx, just double quote the output values like in commit 57d8191569821f07d8652effcccab0b04aa935a0 note that we never use single quotes to quote the output values, I'd prefer to be consistent with current behavior.

I also don't think it's reasonable to expect support for embedding `"` in a path, so that's not a case that I would consider.

ben-grande:
> I don't think we need to resort to shelx, just double quote the output values like in commit [57d8191](https://github.com/pykickstart/pykickstart/commit/57d8191569821f07d8652effcccab0b04aa935a0) note that we never use single quotes to quote the output values, I'd prefer to be consistent with current behavior.

If you want to be consistent, would you rather that I do shlex.quote everything that is already weakly quoted? Because the current behavior is not safe, it breaks installs when the user downloads the ISO to a path that we can't control.

bcl:
No. I want to make the minimum change needed to fix problems. Currently we use `"` as our quotes in the kickstart output so we should stick with that.

tests/commands/harddrive.py
```diff
@@ -52,33 +52,44 @@ def runTest(self):
...
+        line4 = [line[len("harddrive "):] for line in str(data4).splitlines() if line.startswith("harddrive ")][0]
+        assert expected_dir in line4
+
 
 class FC3_TestCase(CommandTest):
     def runTest(self):
         # pass
-        self.assert_parse("harddrive --dir=/install --biospart=part", "harddrive --dir=/install --biospart=part\n")
-        self.assert_parse("harddrive --dir=/install --partition=part", "harddrive --dir=/install --partition=part\n")
+        self.assert_parse("harddrive --dir=\"/install\" --biospart=part", "harddrive --dir=\"/install\" --biospart=part\n")
```
brlane-rht: Could you also test that inputs without quotes result in output with quotes? Since this is a change we need to be sure that previously working inputs still parse correctly.

tests/commands/nfs.py
```diff
@@ -49,22 +49,22 @@ def runTest(self):
 class FC3_TestCase(CommandTest):
     def runTest(self):
         # pass
-        self.assert_parse("nfs --server=1.2.3.4 --dir=/install", "nfs --server=1.2.3.4 --dir=/install\n")
+        self.assert_parse("nfs --server=1.2.3.4 --dir=\"/install\"", "nfs --server=1.2.3.4 --dir=\"/install\"\n")
```
brlane-rht: Same as with harddrive, add a test with the input unquoted and the output quoted.

brlane-rht (changes requested):
Looks pretty good, just a couple small requests. You could also change to using single quoted strings to avoid all the `\"` escaping, but that's up to you.

tests/commands/nfs.py
ben-grande: Done.

tests/commands/harddrive.py
ben-grande: Done.

brlane-rht (approved):
Thanks!

PR #568, opened 2026-04-17
bootc: Add bootloader and composefs arguments

### LINKED ISSUE #567: bootc: Add support for more options (--bootloader, --composefs-backend, --uki-addon)
For the upcoming support in bootc for sealed bootable container images, we will need support in kickstart for the following options:

- `--bootloader` (`grub`, `systemd`, `none`)
- `--composefs-backend` (boolean)
- and likely also the `--uki-addon` (repeatable, string), but it's not "stable" yet.

See details in https://bootc.dev/bootc/man/bootc-install-to-filesystem.8.html.

brlane-rht:
Use --bootloader=[none|grub|systemd] to select bootloader. Use --composefs to use the composefs backend instead of ostree.

Also includes tests.

Fixes #567

### DIFF
--- a/pykickstart/commands/bootc.py
+++ b/pykickstart/commands/bootc.py
@@ -15,7 +15,7 @@
 # License and may only be used or replicated with the express permission of
 # Red Hat, Inc.
 #
-from pykickstart.version import F43
+from pykickstart.version import F43, F45
 from pykickstart.base import KickstartCommand
 from pykickstart.options import KSOptionParser
 
@@ -102,3 +102,37 @@ class F43_Bootc(KickstartCommand):
 
         return self
 
+class F45_Bootc(F43_Bootc):
+    removedKeywords = F43_Bootc.removedKeywords
+    removedAttrs = F43_Bootc.removedAttrs
+
+    def __init__(self, *args, **kwargs):
+        F43_Bootc.__init__(self, *args, **kwargs)
+        self.bootloader = kwargs.get("bootloader", None)
+        self.composefs = kwargs.get("composefs", False)
+
+    def _getArgsAsStr(self):
+        retval = F43_Bootc._getArgsAsStr(self)
+
+        if self.bootloader:
+            retval += ' --bootloader="%s"' % self.bootloader
+        if self.composefs:
+            retval += " --composefs"
+
+        return retval
+
+    def _getParser(self):
+        op = F43_Bootc._getParser(self)
+        op.add_argument("--bootloader",
+                        version=F45,
+                        help="""
+                        Bootloader selection: none, grub, systemd
+                        """)
+        op.add_argument("--composefs",
+                        version=F45,
+                        action="store_true",
+                        default=False,
+                        help="""
+                        Use the composefs backend instead of the the default which is ostree.
+                        """)
+        return op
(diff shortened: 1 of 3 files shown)


pykickstart/commands/bootc.py
```diff
@@ -102,3 +102,37 @@ def parse(self, args):
...
+            retval += " --composefs"
+
+        return retval
+
+    def _getParser(self):
+        op = F43_Bootc._getParser(self)
+        op.add_argument("--bootloader",
+                        version=F45,
+                        help="""
+                        Bootloader selection: grub, systemd, none
```
elkoniu: I would keep the same order here as in commit message: none, grub, systemd.
We could also highlight that "none" is the default.

pykickstart/commands/bootc.py
```diff
@@ -102,3 +102,37 @@ def parse(self, args):
...
+                        """)
+        op.add_argument("--composefs",
+                        version=F45,
+                        action="store_true",
+                        default=False,
+                        help="""
+                        Use composefs backend when true. ostree when false.
```
elkoniu: Wondering if it is worth to mark here that "ostree" is the default (composefs == false).

tests/commands/bootc.py
```diff
@@ -81,3 +81,23 @@ def runTest(self):
...
+
+        cmdstr = "bootc --source-imgref=\"quay.io/centos-bootc/centos-bootc:stream9\" --bootloader=grub"
+        cmdstr_expected = "bootc --stateroot=\"default\" --source-imgref=\"quay.io/centos-bootc/centos-bootc:stream9\" --target-imgref=\"quay.io/centos-bootc/centos-bootc:stream9\" --bootloader=\"grub\"" + "\n"
+        self.assert_parse(cmdstr, cmdstr_expected)
+
+        cmdstr = "bootc --source-imgref=\"quay.io/centos-bootc/centos-bootc:stream9\" --composefs"
+        cmdstr_expected = "bootc --stateroot=\"default\" --source-imgref=\"quay.io/centos-bootc/centos-bootc:stream9\" --target-imgref=\"quay.io/centos-bootc/centos-bootc:stream9\" --composefs" + "\n"
+        self.assert_parse(cmdstr, cmdstr_expected)
+
+        cmdstr = "bootc --source-imgref=\"quay.io/centos-bootc/centos-bootc:stream9\" --bootloader=grub --composefs"
```
elkoniu, line 101: Maybe here "systemd" instead of "grub" to have a more "rich" coverage :)

elkoniu (commented):
Generally LGTM, just few minor comments.

pykickstart/commands/bootc.py
brlane-rht: Yeah, I'm not sure about 'none' right now I'm treating them all as strings so that not settings this excludes it from the output. Asked about that in #567

pykickstart/commands/bootc.py
brlane-rht: oops, I didn't update the help string (I initially had it taking an argument). I'll fix that.

tests/commands/bootc.py
brlane-rht: At this point they're just strings, so it doesn't matter.

elkoniu (approved):
LGTM

M4rtinK (approved):
Looks good to me, thanks! :)

PR #571, opened 2026-04-28
Remove commands that have been deprecated

brlane-rht:


tests/commands/vnc.py
```diff
@@ -76,5 +76,13 @@ def runTest(self):
         parser = self.getParser("vnc")
         self.assertEqual(issubclass(parser.__class__, DeprecatedCommand), True)
 
+
+class F45_TestCase(F43_TestCase):
+    def runTest(self):
+        # make sure the vnc command has been deprecated on Fedora 43
```
rvykydal: minor: seems like obsolete comment

rvykydal:
Looks good to me but I'll rather check that we are OK removing the commands. I think removing means that the kickstart installations with the commands will start failing.

rvykydal (approved):
Looks good to me, thank you.

PR #572, opened 2026-04-28
Fix RHEL document handling

brlane-rht:
The RHEL docs section was still pointing to RHEL9, and standalone pages were difficult to distinguish. This switches to using RHEL10 in the main document, moving RHEL9 to a standalone page, and adds the release version to the title string.


rvykydal (approved):
Looks good to me. Thank you.

PR #573, opened 2026-05-18
Add support fort --type=anchor to %certificate section

rvykydal:
Required for https://github.com/rhinstaller/anaconda/pull/7068

### DIFF
--- a/pykickstart/sections.py
+++ b/pykickstart/sections.py
@@ -36,7 +36,7 @@ from pykickstart.constants import KS_SCRIPT_PRE, KS_SCRIPT_POST, KS_SCRIPT_TRACE
                                   KS_BROKEN_IGNORE, KS_BROKEN_REPORT
 from pykickstart.errors import KickstartParseError, KickstartDeprecationWarning
 from pykickstart.options import KSOptionParser
-from pykickstart.version import FC4, F7, F9, F18, F21, F22, F24, F32, F34, F40, F42
+from pykickstart.version import FC4, F7, F9, F18, F21, F22, F24, F32, F34, F40, F42, F45
 from pykickstart.version import isRHEL, RHEL6, RHEL7, RHEL9, RHEL10
 from pykickstart.i18n import _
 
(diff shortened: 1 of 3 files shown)


pykickstart/sections.py
```diff
@@ -868,10 +869,18 @@ def _getParser(self):
                             The %certificate section is used to specify certificates to be
                             installed into the installer environment and the installed system.
 
-                            The certificate content should be in a Base64 ASCII enconding
+                            The certificate content should be in a Base64 ASCII encoding
                             format. It will be written into a file specified by the
                             --dir and --filename options.
 
+                            The --type option can be used to specify the type of the
```
brlane-rht, line 876: What other types are there, and/or where are they documented?

pykickstart/sections.py
```diff
@@ -917,8 +917,9 @@ def _getParser(self):
                         The directory where the certificate should be installed.
                         Required if --type is not specified.""")
 
-        op.add_argument("--type", dest="type", required=False, version=F45, help="""
-                        The type of the certificate. Supported values: anchor.
+        op.add_argument("--type", dest="type", required=False, version=F45,
+                        choices=["anchor"], help="""
```
bcl: That's fine if this isn't expected to change much, but if anaconda is planning on expanding the list then it's better to just pass it through and let anaconda handle errors.

Also, if this is going to be the only supported type it may make more sense to use `--anchor` instead of `--type` since that implies other options.

bcl (approved):
I'm ok with it as is, but have a couple comments to consider.

pykickstart/sections.py
rvykydal: This doc is the only upstream documentation we have for the %certificate section. So there is only the one type currently.
I'll update it o be clear on this.

pykickstart/sections.py
rvykydal: You are right, let's do the check in anaconda.
It would be also more futureproof WRT to added supported types in Fedora versions.

pykickstart/sections.py
rvykydal: Done

pykickstart/sections.py
rvykydal: Done

brlane-rht (approved):
Looks good, thanks!

PR #574, opened 2026-06-03
Enable Packit COPR builds for Fedora rawhide

KKoukiou:
Add copr_build jobs on pull requests and commits to main. Persistent repo: @rhinstaller/Pykickstart [1]

[1] https://copr.fedorainfracloud.org/coprs/g/rhinstaller/Pykickstart/

### DIFF
--- a/packit.yaml
+++ b/packit.yaml
@@ -1,7 +1,36 @@
-specfile_path: pykickstart.spec.in
-synced_files:
-  - packit.yaml
-  - pykickstart.spec.in
-upstream_project_name: pykickstart
+specfile_path: pykickstart.spec
+upstream_package_name: pykickstart
 downstream_package_name: pykickstart
-dist_git_url: https://src.fedoraproject.org/rpms/pykickstart.git
+
+srpm_build_deps:
+  - git
+  - gettext
+  - make
+  - python3-devel
+  - python3-pip
+  - python3-setuptools
+  - python3-build
+  - python3-sphinx
+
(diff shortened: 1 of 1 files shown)


packit.yaml
```diff
@@ -1,7 +1,36 @@
-specfile_path: pykickstart.spec.in
-synced_files:
-  - packit.yaml
```
KKoukiou, line 3: These are by default synced.

brlane-rht:
I don't use packit for anything (and actually forgot this file was here), so I'd like someone else to review this.
It looks like this has never been working? I don't see any builds in COPR, so what is it going to be used for? I'd be more inclined to just remove it.

KKoukiou:
@brlane-rht in Anaconda we often implement a feature that is dependent on some work from pykickstart. With the current CI setup, we need to wait that pykickstart releases and the feature is available in Rawhide before our side of the feature in Anaconda can get a proper CI test run and get merged. We basically don't have an easy way to do cross project testing in CI between anaconda and pykickstart before the Fedora releases are in place. 
Copr builds make this simpler, we basically plan to add pykickstart copr build from the master branch when building the daily ISO that we are testing. This way we will run our CI tests against the merged code on master branch, and not need to wait for a Fedora release.

Similarly with copr builds from packit on PRs, we can implement CI scenario for open PRs even.

rvykydal (approved):
Would be very helpful for building testing anaconda boot.isos.

PR #577, opened 2026-06-08
Fix filename typo in condition

kallies:
The condition if the resolv.conf exists in the sysimage path should check for the correct filename


brlane-rht (approved):
Thanks!

PR #579, opened 2026-06-08
spec: do not depend on pytest-cov, nor pytest-xdist on RHEL

yselkowitz:
Coverage tests are not run during the package build, in line with Fedora
packaging policy.  xdist is not included in RHEL, so the tests need to be
run serially.

Related: https://src.fedoraproject.org/rpms/pykickstart/pull-request/7


brlane-rht (approved):
Thanks!

PR #584, opened 2026-07-10
Add support for stratis to autopart

vojtechtrefny:
First step for my [Fedora 45 change](https://fedoraproject.org/wiki/Changes/StratisAnacondaSupport). Stratis-specific commands for manual partitioning will follow later.


pykickstart/commands/autopart.py
```diff
@@ -665,3 +665,51 @@ def parse(self, args):
...
+                    .. versionchanged:: %s
+
+                    Partitioning scheme 'stratis' was added.""" % versionToLongString(F45)
+        return op
+
+    def parse(self, args):
+        retval = super().parse(args)
+
+        if self._typeAsStr() == "stratis":
+            incompatible = []
```
brlane-rht, line 689: What options does `--type=stratis` support? I think it would be easier to maintain if this was flipped to just check for compatible options and exclude all others.

vojtechtrefny: `--encrypted`, `--passphrase`, `--nohome`, `--noboot`, `--noswap` and `--hibernation` are supported. So it's 6 supported and 10 unsupported right now.

I agree it would be better for future proofing this, but I don't see a simple way of flipping the logic here -- any idea how to simply check that only compatible options are set? Checking this way is quite simple, because everything is already parsed, flipping the check would mean going through the options with `getParser` again, right?

brlane-rht: You're right, I was thinking we had a dict of them to check but we don't. Maybe in the future we can add something like `conflictingArgs` similar to `conflictingCommands` but for now this is fine.

brlane-rht (approved):
~Looks good!~

brlane-rht (changes requested):
Sorry, could you add a bit more to the commit message. Supported options and maybe a link to an overview of what Stratis is?

vojtechtrefny:
> Sorry, could you add a bit more to the commit message. Supported options and maybe a link to an overview of what Stratis is?

Ok, I've added some information to the commit message.

brlane-rht (approved):
Looks good!

PR #585, opened 2026-07-29
Add commands for manual partitioning for Stratis

vojtechtrefny:
This extends support for creating Stratis storage devices, which was introduced in 5c129c62 for autopart.
Two new commands are added:
- stratispool for creating Stratis pool devices from one or more block devices (partitions) and
- stratisfs for creating Stratis filesystems in the pool.


pykickstart/commands/stratispool.py
```diff
@@ -0,0 +1,165 @@
...
+
+        return self.name == y.name
+
+    def __ne__(self, y):
+        return not self == y
+
+    def _getArgsAsStr(self):
+        retval = ""
+        if self.preexist:
+            retval += " --useexisting"
```
brlane-rht, line 54: Doesn't match the kwargs.get in the __init__ :)

pykickstart/commands/stratisfs.py
```diff
@@ -0,0 +1,160 @@
...
+
+    def _getArgsAsStr(self):
+        retval = ""
+
+        if self.size:
+            retval += " --size=%d" % self.size
+        if self.grow:
+            retval += " --grow"
+        if self.maxSizeMB:
+            retval += " --maxsize=%d" % self.maxSizeMB
```
brlane-rht, line 60: Another couple mismatches with kwargs.get

brlane-rht:
Looks like this is headed in the right direction. I'd also like to see tests for multiple pools and multiple filesystems, you have tests for duplicate names but not for different names.

pykickstart/commands/stratispool.py
vojtechtrefny: Sorry, I don't understand, `preexist` is used in both:
```
self.preexist = kwargs.get("preexist", False)
```
Or do you mean that the kickstart argument is called `useexisting` and the kwarg here is called `preexist`? That's what the other commands do (I basically copied everything from `volgroup` and `logvol` here). I personally think it would make more sense to call both `useexisting`, but changing that only for stratis would be very confusing in Anaconda.

brlane-rht: Ah ok. Slightly confusing, but I guess consistency is better than changing it.

brlane-rht (approved):
Looks good!

